PR #180, opened 2026-01-14
feat(style): add anti-numbering hooks

bendrucker:
## Summary

- Add new `style` plugin with hooks to detect inappropriate numbered sequences in code and documentation
- Detects numbered headers (`# 1. Something`, `## Step 2:`), step/phase comments, and numbered identifiers
- Write tool calls are blocked with a clear message; Edit tool calls ask for confirmation to allow editing existing documents

## Test plan

- [x] Run `shellspec plugins/style/spec.sh` - 15 tests pass
- [ ] Test with a Write that introduces `# 1. First Step` - should be blocked
- [ ] Test with an Edit that adds step comments - should prompt for confirmation

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -84,6 +84,11 @@
       "source": "./plugins/newline"
     },
     {
+      "name": "style",
+      "description": "Code style enforcement hooks",
+      "source": "./plugins/style"
+    },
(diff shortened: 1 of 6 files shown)


bendrucker:
Superseded by #184, which fixes the broken grep pipe in numbered identifier detection.

PR #638, opened 2026-05-24
writing: wordlist refactor and expanded trope patterns

bendrucker:
Moves single-word and phrase rules out of inline regex in `tropes.ts` into line-delimited files under `plugins/writing/wordlists/`. Uses a Porter stemmer for vocabulary matching so inflected forms are handled by a language tool rather than hand-rolled regex suffixes. Scopes conversational patterns (sycophantic openers, "let me" preambles, permission-seeking) to side-effect tools where they're relevant, rather than firing on file writes. Adds a weighted accumulator so borderline marketing verbs can stack into a context reminder without binning each one into a single tier.

## Motivation

This is part of a stack working from a single observation (the assistant kept writing "reaching for `Bash`") into a broader question: which AI-typical patterns am I letting through, and how do I keep the answer fresh as models change?

Three parallel investigations ran in the same session that produced this PR:

- **Corrections mining** over my session history (25 confirmed corrections, 12 coverage gaps).
- **N-gram analysis** comparing assistant text to my own writing (top tells: the `let me <verb>` cluster, sycophantic openers, `you're right`, `want me to`).
- **Prior-art survey** of community AI-trope wordlists and detectors (`sam-paech/slop-forensics`, `NousResearch/autonovel`, slop-score regex pipeline).

Reports are in `tmp/trope-corrections.md`, `tmp/trope-ngrams.md`, `tmp/trope-prior-art.md`. They surfaced ~30 new high-signal phrases plus design questions that reshaped the approach during review.

### Wordlist Files

```
plugins/writing/wordlists/
  vocabulary.txt        # plain deny: single hit blocks
  openers.txt           # plain deny: sentence-start sycophancy
  let-me-verbs.txt      # plain deny: compiled into one cluster regex
  marketing-verbs.txt   # weighted: hits accumulate into a score
```

Each file is line-delimited. `#` starts a comment. Blank lines are ignored.

### Porter Stemmer for Vocabulary Matching

Single-word vocabulary and weighted marketing verbs use a Porter stemmer (`stemmer` npm package) instead of hand-rolled regex suffix groups. Wordlist files contain plain base words (`meticulous`, `empower`, `garner`). At load time, each entry is stemmed and stored in a `Set`. At scan time, input text is tokenized and each word is stemmed before lookup.

This means `meticulously`, `garnered`, `fostering`, `empowered`, `streamlining` are all caught without maintaining explicit suffix patterns. Adding a new word is one line in a text file. No regex knowledge required.

Openers and let-me-verbs stay as compiled regex because their matching is structural (line-start anchoring, "let me" prefix context), not lexical.

### Conversational vs. Document Scoping

Patterns like `let me <verb>`, `you're right`, `want me to`, `would you like`, and `I understand` are assistant conversational tells. They show up when the assistant is composing messages, not when it's writing files or documentation.

These are now marked `sideEffectOnly: true` and only fire on Bash/MCP tools (Slack messages, PR body content, issue comments). File Write/Edit operations skip them. This reduces noise on the most common hook path (writing code and docs) while still catching pasted assistant prose in outbound communications.

#### Deny Tier (Single Hit Blocks or Reminds)

| Pattern | Source | Scope |
|---|---|---|
| Expanded vocabulary (from `slop-forensics`) | Wordlist: `vocabulary.txt` | All |
| Sycophantic openers (`Perfect.`, `Excellent.`, etc.) | Wordlist: `openers.txt` | Side-effect only |
| `you're (absolutely) right` | Inline regex | Side-effect only |
| `want me to <verb>` | Inline regex | Side-effect only |
| `would you like` | Inline regex | Side-effect only |
| `let me <verb>` cluster | Wordlist: `let-me-verbs.txt` | Side-effect only |
| `reaching for` (branch namesake) | Inline regex | All |

#### Context Tier (Soft Reminder Only)

| Pattern | Source |
|---|---|
| Cross-sentence `It isn't X. It is Y.` | Ported from `sam-paech/slop-score` |
| Passive PR summary (`is added`, `was updated`) | Inline regex, prose-only |
| `Tests cover` / `Tests verify` | Inline regex |
| `**path**: description` bullet | Inline regex, prose-only |
| Trailing hedge adverbs (`regardless.`, `nonetheless.`) | Inline regex, prose-only |
| `**Label:**` outside code | Inline regex, prose-only |
| `dig` / `dive` / `wade into` | Inline regex |
| Hedging observation (`looks like`, `appears to`) | Inline regex |
| `I understand` | Inline regex, side-effect only |

#### Weighted (Marketing Verbs)

The hook accumulates `count * weight` per stemmed match and emits a context reminder when introduced text crosses 3.0 cumulative weight.

| Entry | Weight | Rationale |
|---|---|---|
| `empower` | 2.5 | Highly promotional, rarely legitimate |
| `streamline` | 2.5 | Same |
| `generalize` | 2.5 | Vague as a PR summary verb |
| `unlock` (figurative) | 2.0 | |
| `elevate` | 2.0 | |
| `transform` (figurative) | 1.5 | |
| `enhance` | 1.5 | |
| `optimize` | 1.2 | |
| `enable` | 0.8 | Often technically correct |
| `simplify` | 0.8 | Often technically correct |

Two high-weight verbs trip the threshold. `enable` + `simplify` alone will not. `enable` + `empower` will (0.8 + 2.5 > 3.0).

## Design Decisions

<details>
<summary>Stemmer Over Hand-Rolled Suffixes</summary>

The original implementation used a custom suffix-group syntax (`meticulous(ly)`, `empower(s|ed|ing)`) that compiled to regex alternations. During review, the question came up: why maintain inflection patterns by hand when a stemmer handles this generically?

The Porter stemmer (`stemmer` npm, pure JS, zero deps, Bun auto-installs) normalizes both the wordlist entries and the scanned text to stems. This is the standard approach for this kind of matching and eliminates an entire class of maintenance burden. The wordlist files are now plain word lists that anyone can edit without knowing regex.

Structural patterns (openers at line start, "let me" + verb) stay as compiled regex because the match depends on position and context, not just word identity.

</details>

<details>
<summary>Scoping Conversational Patterns to Side-Effect Tools</summary>

During review, the question was raised: are patterns like `let me check` and `you're right` realistic in documents? An assistant wouldn't write "let me check" into a `.md` file. These are conversational tells that surface in Slack messages, PR review comments, and similar outbound communications composed by the assistant.

Scoping them to `sideEffectOnly` reduces false-positive noise on the most common hook path (file editing) while still catching pasted assistant prose in communications. If a pattern later proves relevant to file writes too, removing the flag is a one-line change.

</details>

<details>
<summary>Per-Word Weights Instead of Tier Buckets for Marketing Verbs</summary>

The original design proposed three tiers (deny / context / watch) for every word. During the design walkthrough:

> "The concept of tiers seems questionable. We should be able to score words based on how marketing-centric they are. We shouldn't just be grouping things into lists and treating them all at the same level."

That reshaped marketing verbs into a weighted wordlist with an accumulator. `enable` and `empower` are not equivalent and the design now reflects that.

</details>

<details>
<summary>Constrained Wordlist Size</summary>

From the prior-art survey we could have copied the full ~2000-word `slop-forensics` list. The PR picks roughly 20 distinctive additions instead. High-signal, hand-picked entries outperform exhaustive lists both on false-positive rate and on context cost (the wordlists compile into matchers that scan every Write/Edit). The `writing:analyze` skill (separate PR) is the mechanism for growing the list over time based on evidence, not bulk import.

</details>

### Hook Skip for Wordlist Files

`check-tropes.ts` skips Write/Edit operations whose path matches `/wordlists/*.txt`, since these files necessarily contain the flagged vocabulary they catalog.

## Test Coverage

<details>
<summary>What Is Tested</summary>

- `wordlists.test.ts`: stemmed matching (base word, inflected forms, counts), weighted stem accumulation (threshold behavior, inflected matching, multiple occurrences), plain wordlist compilation, comments/blank-line handling.
- `tropes.test.ts`: each new pattern has positive and negative cases. New `sideEffectOnly scoping` section verifies all conversational patterns fire in sideEffect context and skip in file context. Non-conversational patterns still fire in file context.
- `check-tropes.test.ts`: wordlist-file skip, sycophantic opener correctly skipped in file Edit (sideEffectOnly), diff-aware filtering for new patterns.

</details>

## Items Worth a Careful Look

1. **Stemmer precision.** Porter stemming can collide unrelated words (e.g., `universal` and `universe` share a stem). For a curated ~40-word list, collision risk is low. If a false positive surfaces, the fix is removing the offending base word, not fixing the stemmer.
2. **Marketing-verb threshold (3.0).** Starting calibration. Real tuning happens via PR B's scoring command.
3. **Cross-sentence "not X" regex** anchors on a small pronoun set to keep false positives down. Named-subject variants deferred.
4. **`let me <verb>` deny** fires on sideEffect tools. This catches assistant prose in Slack/PR bodies but won't trigger when writing files.

## Deferred to PR B (Separate Session)

- `writing:score <file>` command for per-1000-token scoring of arbitrary prose (review use case).
- Code-comment extraction so prose-only patterns scan `//`, `#`, `/* */` blocks in non-prose files. Addresses the substitution-drift case (em dashes were banned, semicolons relocated into code comments).
- User-level and project-level wordlist overrides (`~/.claude/wordlists/`, `.claude/wordlists/`).
- Remark-based prose extraction from markdown (replace regex `stripCode` with AST-based stripping).

## Related Work

- #636 (merged): session-plugin queries this PR's analysis sources lean on.
- #639 (pending): `writing:analyze` skill that curates the wordlists this PR adds.
- #640 (pending): `CLAUDE.md` curation principle that motivates the constrained wordlist size.
- Branch `session-cross-machine-design`: design doc for multi-machine session corpus.

## References (Working Notes)

- `tmp/design-trope-overhaul.md`: full design proposal and resolved decisions.
- `tmp/trope-corrections.md`: 25 corrections from session history, ranked coverage gaps.
- `tmp/trope-ngrams.md`: corpus stats, top phrases.
- `tmp/trope-prior-art.md`: wordlist source links, scoring formulas, anti-patterns.


plugins/writing/skills/writing/SKILL.md
```diff
@@ -11,42 +11,97 @@ user-invocable: false
...
 Never use spaced em dashes (` — `). This is the most recognizable AI writing tell.
 
 - Use unspaced em dashes: `word—word`
 - Or rewrite with commas, colons, or parentheses
 
-### Vocabulary
+#### Vocabulary
 
-These words are strongly associated with AI-generated text. Do not use them:
+These words are strongly associated with AI-generated text. The full list lives in [`wordlists/vocabulary.txt`](../../wordlists/vocabulary.txt). Matching uses a Porter stemmer, so inflected forms (`meticulously`, `garnered`, `fostering`) are caught automatically from base entries. Highlights:
```
bendrucker: live in
too verbose, no need for stemming examples

plugins/writing/skills/writing/SKILL.md
```diff
@@ -11,42 +11,97 @@ user-invocable: false
...
 
-- delve, tapestry, landscape (figurative), meticulous/meticulously
+- delve, tapestry, landscape (figurative), meticulous
```
bendrucker: repeating all these is not necessary and very likely even harmful, I'd sooner remove all of them

plugins/writing/skills/writing/SKILL.md
bendrucker, line 1: too much is being added here, keep SKILL.md compact

PR #642, opened 2026-05-24
writing: rewrite skill, analyze fixes, wordlist curation

bendrucker:
Extends the writing plugin: a user-invoked `writing:rewrite` skill, fixes to `writing:analyze` so its curation is trustworthy and its candidate miner scans the surface the hook actually enforces, and a round of wordlist curation against a larger combined corpus that both prunes stale entries and adds PR-body tells.

## Rewrite skill

A user-invocable `writing:rewrite` skill that rewrites input text in direct, concise style. It accepts text from a file path, inline arguments, or the clipboard, and detects violations programmatically rather than loading wordlists into context.

#### Lint script (`scripts/lint.ts`)

A programmatic linter that intersects input text against the writing plugin's vocabulary, marketing verbs, and inline patterns (em dashes, copula avoidance, hedging). It reuses the Porter stemmer from `hooks/wordlists.ts` for vocabulary matching and outputs one violation per line as `line:col: category: message`, exiting non-zero when it finds any. The agent runs it before and after rewriting.

#### Skill definition (`SKILL.md`)

Input resolution (file path, inline text, clipboard via `pbpaste`), instructions to run the lint script, and structural rewriting guidance. The skill runs in `context: fork` so it operates in a sidechain.

#### Style rules (`references/style-rules.md`)

Trimmed to structural rules only: sentence structure, semicolons, filler phrases, tone. Anything the lint script detects programmatically is gone from the reference file.

<details>
<summary>Why programmatic linting instead of loading the wordlists</summary>

Having the agent read `wordlists/*.txt` to learn what to avoid wastes tokens on a full vocabulary dump and shows the model the exact words we want it to avoid, which risks steering it toward them. The lint script runs the same stemmer-based matching as the hook but reports only the violations present in the input, so the agent sees `line 3: vocabulary: 'delve' is AI-typical` and fixes it without ever reading the list.
</details>

## Analyze fixes

Three changes so `writing:analyze` recommends the right edits against the surface the hook enforces.

#### Keep/remove no longer uses lift

A rule is kept when the model uses it more per token than the user and it clears a `--min-count` floor (new flag, default 5). Otherwise the report removes it as **dead** (too rare to fire) or **not distinctive** (the user uses it as often). Lift drove this before, but the human baseline is small, so Laplace smoothing puts a high floor on the lift denominator: any word the user never types reads as low lift regardless of how often the model writes it. That gate flagged `delve`, `Perfect`, and `robust` for removal. The direct rate comparison keeps them.

#### Structural audit sourced from the hook

`structural.ts` now imports the hook's `PATTERNS` from `hooks/tropes.ts` and filters to the non-wordlist regexes, instead of a hand-copied list that had already drifted and under-reported. The audit can no longer diverge from what the hook enforces.

#### Candidate miner scoped to deliverable prose

The n-gram candidate miner now reads only deliverable prose (file writes and Bash commit/PR/MR bodies), not conversational assistant text. The hook scans deliverables and side-effect inputs, never chat, so mining chat flooded the candidate list with narration ("now let me", "let me check") at high lift that no rule can act on.

## Wordlist curation

Re-runs the corrected audit against a larger combined corpus (cross-machine history merged in #662) and curates in both directions.

#### Prune and refine

The earlier prune (53 entries to 11) holds against the larger baseline: every surviving entry audits as distinctive. Two refinements fell out. `Perfect` left the openers list because the user says it as an acknowledgment about as often as the model. In the marketing verbs, `simplify` left (normal engineering vocabulary the user writes more than the model) and `enhance`/`streamline` came back once the larger corpus showed the model still produces them.

#### PR-body tells

Adds tells found by comparing the model's PR bodies against the user's hand-written, pre-AI pull requests, where each phrase appears zero times:

- `flowery-phrases.txt`: distinctive multi-word phrasing (`source of truth`, `escape hatch`, `self-sufficient`, `fail loudly`). Matching stems each word and looks for the stemmed token sequence, so inflection, hyphenation, and case are caught (`sources of truth`, `source-of-truth`, `fails loudly`). A context-tier reminder fires on the first match.
- `soft-phrasing.txt`: `cleanly`, weighted so it fires only when it stacks rather than on a single legitimate use.

Both rules are `fileOnly`, so code is never flagged. The verbs `owns`/`carries` were considered and rejected: the bare stem also matches the possessive "its own", which the baseline uses, so the rule would flag the user's own voice. Matching is a plain stemmed subsequence scan, which benchmarks at full recall with no false positives. A word-interior n-gram prefilter is documented as the scaling lever for if a mined list ever reaches thousands of entries.

#### Known limitation

The audit measures `text_content` (chat), so it cannot see tells that live only in deliverables. It reports the new phrase tells as `dead` (zero chat hits). The methodology reference notes this. Auditing wordlist health against the deliverable corpus directly is follow-up work.

## Test plan

- [ ] `echo "delve into the intricacies" | bun plugins/writing/skills/rewrite/scripts/lint.ts` reports violations
- [ ] Clean text exits 0 with "No violations found."
- [ ] Code blocks are excluded from scanning
- [ ] `/writing:rewrite` with inline text runs the linter, fixes violations, and re-lints clean
- [ ] `/writing:rewrite path/to/file.md` reads and rewrites file contents
- [ ] `/writing:rewrite` with no args reads from the clipboard
- [ ] `bun test plugins/writing` exercises the removal logic, pruned wordlists, and new matchers
- [ ] `writing:analyze` proposes no removals against the kept entries

### DIFF
--- a/plugins/writing/skills/rewrite/SKILL.md
+++ b/plugins/writing/skills/rewrite/SKILL.md
@@ -6,6 +6,9 @@ description: >-
   before sending. Strips AI voice, slop, and filler while preserving meaning.
 user-invocable: true
 context: fork
+allowed-tools:
+  - Bash
+  - Read
 ---
 
 # Rewrite
@@ -18,11 +21,11 @@ $ARGUMENTS
 
 Read input from one of these sources, checked in order:
 
-#### File path
+#### File Path
 
 If `$ARGUMENTS` is a path to an existing file, read the file contents.
 
-#### Inline text
+#### Inline Text
 
 If `$ARGUMENTS` contains text, use it directly.
 
@@ -30,16 +33,30 @@ If `$ARGUMENTS` contains text, use it directly.
 
 If `$ARGUMENTS` is empty, read from the clipboard with `pbpaste`.
 
+## Lint
+
+Before and after rewriting, run the lint script to find violations:
+
+```bash
+echo "<text>" | bun ${CLAUDE_SKILL_DIR}/scripts/lint.ts
+```
+
+Or pass a file path directly:
+
+```bash
+bun ${CLAUDE_SKILL_DIR}/scripts/lint.ts path/to/file.md
+```
+
+The script outputs one violation per line (`line:col: category: message`). Fix every reported violation. Re-run until the output is clean.
+
 ## Rewriting
 
-Read the wordlist files referenced in [references/style-rules.md](references/style-rules.md), then apply every rule. Focus on:
+Fix all lint violations, then apply the structural rules from [references/style-rules.md](references/style-rules.md):
 
-- Replacing banned vocabulary and promotional language with natural alternatives
-- Removing filler phrases and hedging
 - Breaking long compound sentences into shorter ones
 - Converting passive voice to active
 - Cutting sentences that restate previous ones
-- Removing sycophantic openers
+- Removing filler phrases and hedging
 
 Do not add, remove, or restructure the content's meaning. The output should convey the same information in fewer, clearer words.
 
--- a/plugins/writing/skills/rewrite/SKILL.md
+++ b/plugins/writing/skills/rewrite/SKILL.md
@@ -1,0 +1,50 @@
+---
+name: writing:rewrite
+description: >-
+  Rewrite text in direct, concise style. Use when you have functional content
+  (an explanation, a draft message, notes) that needs to read as polished prose
+  before sending. Strips AI voice, slop, and filler while preserving meaning.
+user-invocable: true
+context: fork
+---
+
+# Rewrite
+
+Rewrite input text to match the style rules in [references/style-rules.md](references/style-rules.md). Preserve all functional information. Change only voice, word choice, and sentence structure.
+
+## Input
+
+$ARGUMENTS
+
+Read input from one of these sources, checked in order:
+
+#### File path
+
+If `$ARGUMENTS` is a path to an existing file, read the file contents.
+
+#### Inline text
+
+If `$ARGUMENTS` contains text, use it directly.
+
+#### Clipboard
+
+If `$ARGUMENTS` is empty, read from the clipboard with `pbpaste`.
+
+## Rewriting
+
+Apply every rule in [references/style-rules.md](references/style-rules.md). Focus on:
+
+- Replacing banned vocabulary and promotional language with natural alternatives
+- Removing filler phrases and hedging
+- Breaking long compound sentences into shorter ones
+- Converting passive voice to active
+- Cutting sentences that restate previous ones
+- Removing sycophantic openers
+
+Do not add, remove, or restructure the content's meaning. The output should convey the same information in fewer, clearer words.
+
+## Output
+
+Display the rewritten text directly. Do not wrap it in a code block unless the input was code.
+
+If the user asks, copy the result to the clipboard with `pbcopy`.
--- a/plugins/writing/skills/rewrite/SKILL.md
+++ b/plugins/writing/skills/rewrite/SKILL.md
@@ -1,9 +1,10 @@
 ---
 name: writing:rewrite
 description: >-
-  Rewrite text in direct, concise style. Use when you have functional content
-  (an explanation, a draft message, notes) that needs to read as polished prose
-  before sending. Strips AI voice, slop, and filler while preserving meaning.
+  Rewrite a draft message, explanation, or note to strip AI voice, slop, and
+  filler while preserving meaning. Use to polish functional prose before
+  sending.
+argument-hint: "[file path or text; omit to read input from the clipboard]"
 user-invocable: true
 context: fork
 allowed-tools:
@@ -19,7 +20,7 @@ Rewrite input text to match the style rules in [references/style-rules.md](refer
 
 $ARGUMENTS
 
-Read input from one of these sources, checked in order:
+Read input from one of these sources:
 
 #### File Path
 
@@ -29,9 +30,7 @@ If `$ARGUMENTS` is a path to an existing file, read the file contents.
 
 If `$ARGUMENTS` contains text, use it directly.
 
-#### Clipboard
-
-If `$ARGUMENTS` is empty, read from the clipboard with `pbpaste`.
+If `$ARGUMENTS` is empty, ask for the text to rewrite. You may offer to read it from the clipboard.
 
 ## Lint
 
@@ -64,4 +63,4 @@ Do not add, remove, or restructure the content's meaning. The output should conv
 
 Display the rewritten text directly. Do not wrap it in a code block unless the input was code.
 
-If the user asks, copy the result to the clipboard with `pbcopy`.
+If the user asks, copy the result to the clipboard.
--- a/plugins/writing/skills/rewrite/SKILL.md
+++ b/plugins/writing/skills/rewrite/SKILL.md
@@ -32,7 +32,7 @@ If `$ARGUMENTS` is empty, read from the clipboard with `pbpaste`.
 
 ## Rewriting
 
-Apply every rule in [references/style-rules.md](references/style-rules.md). Focus on:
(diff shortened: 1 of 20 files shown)


bendrucker:
* **#642** <a href="https://app.graphite.com/github/pr/bendrucker/claude/642?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#664](https://github.com/bendrucker/claude/pull/664) <a href="https://app.graphite.com/github/pr/bendrucker/claude/664?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#665](https://github.com/bendrucker/claude/pull/665) <a href="https://app.graphite.com/github/pr/bendrucker/claude/665?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/bendrucker/claude/642?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

plugins/writing/skills/rewrite/SKILL.md
```diff
@@ -0,0 +1,67 @@
+---
+name: writing:rewrite
+description: >-
+  Rewrite text in direct, concise style. Use when you have functional content
+  (an explanation, a draft message, notes) that needs to read as polished prose
+  before sending. Strips AI voice, slop, and filler while preserving meaning.
+user-invocable: true
+context: fork
+allowed-tools:
+  - Bash
```
bendrucker, line 11: Seems overly broad.

plugins/writing/skills/rewrite/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Input
+
+$ARGUMENTS
+
+Read input from one of these sources, checked in order:
+
+#### File Path
+
+If `$ARGUMENTS` is a path to an existing file, read the file contents.
```
bendrucker, line 27: We should provide argument hints in the front matter.

plugins/writing/skills/rewrite/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+
+#### Inline Text
+
+If `$ARGUMENTS` contains text, use it directly.
+
+#### Clipboard
+
+If `$ARGUMENTS` is empty, read from the clipboard with `pbpaste`.
```
bendrucker: Not sure it makes sense for this plugin to be macOS-specific nor is it really necessary for this skill to integrate clipboard handling. It could simply mention that it can offer to read from the clipboard shouldnt be a default behavior

plugins/writing/skills/rewrite/scripts/lint.ts
```diff
@@ -0,0 +1,224 @@
+#!/usr/bin/env bun
+import { join } from "node:path";
+import { stemmer } from "stemmer";
+
+const WORDLISTS_DIR = join(import.meta.dirname, "..", "..", "..", "wordlists");
```
bendrucker: smell: Ancestor path traversal is a smell

plugins/writing/wordlists/flowery-phrases.txt
```diff
@@ -0,0 +1,21 @@
+# Flowery multi-word phrases the model reaches for in deliverable prose.
+# One phrase per line. Matching stems each word (Porter) and looks for the
+# stemmed token sequence, so inflection, spacing, and hyphenation are caught
+# automatically. "sources of truth", "source-of-truth", and "Source Of Truth"
+# all match "source of truth". "fails loudly" matches "fail loudly". A
+# context-tier reminder fires on the first occurrence in introduced prose.
+#
+# Added from the hand-written PR-body baseline (writing:analyze). Each phrase
+# appears in the model's PR/MR/commit bodies and zero times across the user's
+# 209 pre-AI pull requests. These are distinctive enough to flag on a single
```
bendrucker: 209 -- useless details  
  
Comment is extremely bloated. Cut it way down.

plugins/writing/wordlists/marketing-verbs.txt
```diff
@@ -4,21 +4,13 @@
 # A context-tier reminder fires when the sum of weighted hits in the
 # introduced text reaches the threshold defined in tropes.ts.
 #
-# Promotional and vague verbs (high weight).
+# Curated against session history by writing:analyze. Every verb here is one the
```
bendrucker: Comment is over-narrating and creating a diary. Cut it down.

plugins/writing/wordlists/soft-phrasing.txt
```diff
@@ -0,0 +1,19 @@
+# Soft phrasing tells, weighted by how strongly they read as filler.
```
bendrucker: Again, edit this down

plugins/writing/wordlists/vocabulary.txt
```diff
@@ -1,43 +1,16 @@
 # Single-word AI vocabulary deny list.
 # One entry per line. Matching uses a Porter stemmer, so inflected forms
-# (e.g., "meticulously", "garnered") are caught automatically.
+# (e.g., "meticulously", "leveraged") are caught automatically.
 #
-# Original deny list.
+# Curated against session history by writing:analyze. Each entry below is one
+# the model uses far more than the user and still produces regularly. Words the
+# model has stopped using (landscape, paradigm, synergy, myriad, ...) were
```
bendrucker: "were removed" don't narrate history

plugins/writing/skills/rewrite/SKILL.md
bendrucker: Tightened the description to the skill's actual scope (stripping AI voice and filler from a draft) instead of the broad framing.

plugins/writing/skills/rewrite/SKILL.md
bendrucker: Added an `argument-hint` to the frontmatter.

plugins/writing/skills/rewrite/SKILL.md
bendrucker: Dropped the macOS-specific clipboard handling. Empty input now asks for the text and only offers to read the clipboard; it is no longer the default, and the skill is platform-neutral.

plugins/writing/skills/rewrite/scripts/lint.ts
bendrucker: `lint.ts` now resolves the wordlists from `CLAUDE_PLUGIN_ROOT`, falling back to the relative path only for direct runs. #665 then folds the lint onto the shared matcher, so the path resolution lives in one place.

plugins/writing/wordlists/flowery-phrases.txt
bendrucker: Cut the header to a single line.

plugins/writing/wordlists/marketing-verbs.txt
bendrucker: Cut the header down and dropped the diary.

plugins/writing/wordlists/soft-phrasing.txt
bendrucker: Trimmed the header.

plugins/writing/wordlists/vocabulary.txt
bendrucker: Removed the history narration; the comment now just describes what the list is.

bendrucker:
### Merge activity

* **Jun 5, 9:44 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/bendrucker/claude/642).
* **Jun 5, 9:44 PM UTC**: @bendrucker merged this pull request with [Graphite](https://app.graphite.com/github/pr/bendrucker/claude/642).

PR #664, opened 2026-05-29
writing:analyze: deliverable-aware audit + local voice baseline

bendrucker:
Audits `writing:analyze` rules against the surface where the hook actually fires them, so deliverable-only tells stop reading as dead. Adds a local-only voice baseline (the user's hand-written pull requests) as the comparison surface, a corrective-feedback extractor, quote-in-context for every candidate and tell, and a paste filter for the user baseline.

Stacks on #642.

#### The blind spot this fixes

The rule-health audit measured every wordlist entry against chat text. Most rules fire on deliverables, not chat, and the model's chat usage tracks its habit closely enough for chat-surface rules. It breaks for tells that live only in deliverables. The flowery phrases (`source of truth`, `escape hatch`, `self-sufficient`, `fail loudly`) and the soft `cleanly` were curated from PR-body evidence, so the chat audit reported all four flowery phrases as dead and proposed them for removal.

Each rule is now judged on its firing surface. Chat-surface rules (openers, sycophantic patterns, conversational vocabulary) keep the chat comparison. The `fileOnly` deliverable rules compare the model's deliverable-prose rate against the user's voice baseline. A phrase frequent in deliverables and absent from the baseline now reads as keep:

| phrase | before | after | deliverable/M | baseline/M |
| --- | --- | --- | --- | --- |
| `source of truth` | remove (dead) | keep | 169.2 | 0.0 |
| `fail loudly` | remove (dead) | keep | 76.0 | 0.0 |
| `escape hatch` | remove (dead) | keep | 44.1 | 0.0 |
| `self-sufficient` | remove (dead) | keep | 17.2 | 0.0 |
| `cleanly` | keep | keep | 215.7 | 0.0 |

`marketing-verbs.txt` stays on the chat audit because its hook group is not `fileOnly`, so it fires on side-effect inputs too and a deliverable-only count undercounts it.

#### Voice baseline

The user's true deliverable voice is their hand-written, pre-AI pull requests, not their chat. A tell is real if it is frequent in the model's deliverables and absent from this baseline. The baseline is local-only and never committed: it lives outside the repository at the plugin data dir and will grow to include private writing. `ingest-voice.ts` normalizes sources (a `gh search prs` source and the already-present seed corpus) into the data dir. `voice-profile.ts` builds a word n-gram profile from it. A configurable `--self-name` drops pasted prose that refers to the user in the third person from the baseline. Without a profile, deliverable rules are still measured on the deliverable corpus (never the chat audit, which cannot match a multi-word phrase) and an alive rule reads `keep (no baseline)` rather than being proposed for removal.

#### Also in this change

- Corrective-feedback extractor: short, human-authored messages that name a writing problem (a frustration lexicon), paired with the preceding model output. These are higher-signal than the inferred long-assistant/short-user correction candidates.
- Quote-in-context: every candidate and audited deliverable tell ships with a context window and a source pointer, so a curation decision is spot-checkable.
- Paste-aware user baseline: a heuristic drops human-pasted model output (very long messages, third-person self-reference, document structure) so the model's phrasing does not appear on both sides of the lift ratio. On the combined index this trims the human user corpus by more than half.

#### Verification

Type checking, the writing test suite, and skill linting all run clean. New unit tests cover the corpus parser, the profile lookup (including the absent-phrase signal), the frustration regex, the deliverable phrase audit, and the quote finder. Each new script was exercised end-to-end against the session index and the existing baseline corpus, including both `ingest-voice` sources. The profile and corpus write only to the local data dir, and no baseline files are staged for commit.

### DIFF
--- a/plugins/writing/skills/analyze/SKILL.md
+++ b/plugins/writing/skills/analyze/SKILL.md
@@ -26,6 +26,19 @@ DB_PATH=$(<session-skill-dir>/scripts/refresh.ts --refresh)
 
 The refresh script prints the resolved DB path to stdout.
 
+Build the local voice baseline once (and refresh it as new writing accumulates). It is the comparison surface for deliverable-aware rule health. The baseline is local-only and never committed; it lives at `CLAUDE_PLUGIN_DATA` (else `~/.claude/plugins/data/writing-bendrucker`).
+
+```bash
+# Seed from the already-present delimited corpus (no re-fetch needed)
+${CLAUDE_SKILL_DIR}/scripts/ingest-voice.ts --source file --file <data-dir>/voice-baseline/github-prs.txt
+# Or fetch fresh merged PRs (designed to add more sources later)
+${CLAUDE_SKILL_DIR}/scripts/ingest-voice.ts --source github --author <user> --created 2019-01-01..2025-01-01
+# Build the profile the audit reads
+${CLAUDE_SKILL_DIR}/scripts/voice-profile.ts
+```
+
+If no profile exists, analyze still runs: deliverable-surface rules fall back to the chat audit and the report flags the baseline as not loaded.
+
 ## Run
 
 Pass the DB path via `--session-db`:
@@ -36,7 +49,7 @@ ${CLAUDE_SKILL_DIR}/scripts/analyze.ts --session-db "$DB_PATH" --since 2026-04-0
 ${CLAUDE_SKILL_DIR}/scripts/analyze.ts --session-db "$DB_PATH" --project bendrucker.me --min-lift 7
 ```
 
-Run with `--help` for all flags. Writes a markdown report to `tmp/trope-analysis-<date>.md` (override with `--out`).
+Run with `--help` for all flags. `--data-dir` overrides where the voice baseline is read from (default: `CLAUDE_PLUGIN_DATA` or `~/.claude/plugins/data/writing-bendrucker`). Writes a markdown report to `tmp/trope-analysis-<date>.md` (override with `--out`). The report may quote any host in the combined index, so keep it under `tmp/` and never paste host-specific content into committed work.
 
 ## Metrics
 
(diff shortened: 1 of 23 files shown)


bendrucker:
* **#664** <a href="https://app.graphite.com/github/pr/bendrucker/claude/664?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/bendrucker/claude/664?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#642** <a href="https://app.graphite.com/github/pr/bendrucker/claude/642?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#665](https://github.com/bendrucker/claude/pull/665) <a href="https://app.graphite.com/github/pr/bendrucker/claude/665?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

plugins/writing/skills/analyze/SKILL.md
```diff
@@ -26,6 +26,19 @@ DB_PATH=$(<session-skill-dir>/scripts/refresh.ts --refresh)
 
 The refresh script prints the resolved DB path to stdout.
 
+Build the local voice baseline once (and refresh it as new writing accumulates). It is the comparison surface for deliverable-aware rule health. The baseline is local-only and never committed; it lives at `CLAUDE_PLUGIN_DATA` (else `~/.claude/plugins/data/writing-bendrucker`).
```
bendrucker: Itself; the semicolon and everything after it is noise, mostly, we should say that it's stored in Claude plugin data. "Lives at" is itself an AI tell.

plugins/writing/skills/analyze/SKILL.md
```diff
@@ -46,24 +59,26 @@ Run with `--help` for all flags. Writes a markdown report to `tmp/trope-analysis
 
 ## Output
 
-- Summary stats (corpus sizes, rule count, model breakdown)
-- Proposed removals, each tagged **dead** (model produced it fewer than `--min-count` times) or **not distinctive** (user uses it at least as often as the model)
-- Proposed additions (high-lift n-grams not already covered)
-- Rule health table (every entry with type and `keep` / `remove (reason)`)
+- Summary stats (corpus sizes, voice-baseline size, rule count, model breakdown)
+- Proposed removals, each tagged **dead** (model produced it fewer than `--min-count` times on its firing surface) or **not distinctive** (baseline uses it at least as often as the model)
```
bendrucker: "its firing surface" AI slop

plugins/writing/skills/analyze/references/methodology.md
```diff
@@ -39,19 +39,53 @@ Installs DuckDB's FTS extension and materializes per-role corpus tables (`fts_as
...
+
+When no voice profile is loaded (the local baseline has not been built), deliverable-surface rules are still measured on the deliverable corpus, not the chat audit. The chat audit cannot score them: it stems each entry and joins against single-word FTS tokens, so a multi-word phrase never matches and would read as a false `dead`. Without a baseline the distinctiveness check is skipped, so an alive rule is reported `keep (no baseline)` rather than proposed for removal. Build the baseline with `ingest-voice.ts` then `voice-profile.ts` (see "Voice baseline" below) to get the full deliverable-aware verdicts.
 
 Because removed single words are no longer in the wordlist, a later audit cannot resurface them (the additions pipeline only mines multi-word n-grams). If the model regresses to a removed word, add it back by hand.
 
+## Voice baseline
```
bendrucker, line 59: Fix heading style, look for opportunities to cut down this document

plugins/writing/skills/analyze/SKILL.md
bendrucker: Dropped "Lives at" and the trailing noise. It now states the baseline is stored in the plugin data directory.

plugins/writing/skills/analyze/SKILL.md
bendrucker: Replaced "its firing surface" with plain wording for where the hook fires.

plugins/writing/skills/analyze/references/methodology.md
bendrucker: Normalized the headings to sentence-case `####` and trimmed the document (about 12%, plus per-line tightening).

PR #704, opened 2026-06-02
hook: centralize hook stdin/stdout plumbing

bendrucker:
Centralizes the stdin/stdout plumbing that every hook hand-rolled around its transform. Each converted hook reduces to a pure `(input) => output | null` function plus a one-line `runHook` call.

## Changes

Adds `@bendrucker/claude-hook` at `packages/hook/`, exporting `runHook(handler, label)`. It reads stdin via the kit's `readStdinJson`, logs `[${label}] Failed to parse hook input: ...` and returns on parse error, awaits the handler, and writes the result with `writeStdoutJson` when truthy. This matches the previous per-hook behavior exactly.

Converted these hooks to export their transform and delegate to `runHook`:

- `linear/default-state`
- `mac/sandbox`
- `type-ignore` (`detect.ts`)
- `issue/approve-read` (also drops a dynamic `import()` of the runners)
- `shortcuts/open`
- `tmux/target` (wrapped in a closure to pass `process.env.TMUX_PANE`)
- `style/numbering` (wrapped in a closure to pass the CLI `mode` arg)
- `style/headings`
- `writing/tropes` (`check-tropes.ts`)
- `git/block-default-branch-commit`

Each plugin declares `@bendrucker/claude-hook` as a workspace dependency. For `type-ignore`, `shortcuts`, `tmux`, and `writing` the direct `@constellos/claude-code-kit` dependency is now unused (the kit comes transitively through the hook package), so I dropped it.

#### Skipped

- `tmux/notification`: a `cleye` CLI with tmux side effects, not a stdin/stdout transform.
- `user/hooks/webfetch-block` and `user/hooks/worktree`: `user/` has no `package.json`, so they cannot declare a workspace dependency. Left on the old pattern per the constraint to skip rather than force-fit.
- The remaining `grep` matches (`scripts/fetch.ts`, `scripts/check.ts`, `newline/scripts/*`, `pull-request/scripts/validate.ts`, etc.) are CLI scripts, not hook entry points, and do not follow the canonical shape.

## Testing

Added `packages/hook/hook.test.ts` covering the four `runHook` behaviors: an object result writes JSON, `null` writes nothing, async handlers are awaited, and a parse error logs the labeled message and writes nothing. I mocked `@constellos/claude-code-kit/runners` to control `readStdinJson` and capture `writeStdoutJson`, since mocking real stdin under `bun test` is impractical.

Also exercised converted hooks end-to-end with real piped stdin to confirm output and the parse-error path. Verified the existing test suite, `bun run build`, `bun run lint`, and the `check-plugin-imports`, `check-plugin-deps`, `check-hook-quoting`, `check-mcp-matchers`, and `check-marketplace` scripts.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -19,6 +19,12 @@
         "gray-matter": "^4.0.3",
       },
     },
+    "packages/hook": {
+      "name": "@bendrucker/claude-hook",
+      "dependencies": {
+        "@constellos/claude-code-kit": "^0.4.0",
+      },
+    },
     "packages/marketplace": {
       "name": "@bendrucker/claude-marketplace",
     },
@@ -64,6 +70,7 @@
       "name": "git",
       "dependencies": {
         "@anthropic-ai/claude-agent-sdk": "^0.2.25",
+        "@bendrucker/claude-hook": "workspace:*",
         "@constellos/claude-code-kit": "^0.4.0",
       },
     },
@@ -99,6 +106,7 @@
       "name": "issue",
       "dependencies": {
         "@anthropic-ai/claude-agent-sdk": "^0.2.25",
+        "@bendrucker/claude-hook": "workspace:*",
         "cleye": "^1.3.2",
       },
     },
@@ -106,6 +114,7 @@
       "name": "linear",
       "dependencies": {
         "@anthropic-ai/claude-agent-sdk": "^0.2.25",
+        "@bendrucker/claude-hook": "workspace:*",
         "@constellos/claude-code-kit": "^0.4.0",
       },
     },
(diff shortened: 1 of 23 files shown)


bendrucker:
Closing unmerged. This makes distributed plugin hooks depend on a private `workspace:*` package (`@bendrucker/claude-hook`), which Bun's auto-install cannot resolve on an external install, so every converted hook would fail with `Cannot find module`. CI passed only because the monorepo runs `bun install` and links the workspace, masking the break.

Plugin runtime code can depend only on published npm packages (the reason every hook imports `@constellos/claude-code-kit`). Keeping the hooks self-contained instead; the residual boilerplate over the kit's I/O utilities is a few lines and stays installable.

PR #727, opened 2026-06-03
review: surface poll failures and run Monitor as one process

bendrucker:
Hardens the review dashboard's hands-off Monitor loop against two failure modes: poll failures that were silently swallowed, and a shell `while/sleep` loop that breaks under the `Monitor` tool on macOS.

#### Visible Poll Failures

`poll.ts` `fetchUrls` returned an empty array for both an empty queue and a failed source command, so an all-sources-failing sandbox TLS outage was byte-identical to "no new reviews." It now returns a discriminated result (`{ ok: true; urls }` vs `{ ok: false; reason }`). A failed source still contributes zero URLs (one source's outage never stalls the loop), but the caller emits a structured `{"type":"source-error",...}` line so the failure is visible. I removed the `|| true` swallowing from the loop.

#### Single Long-Lived Process

The old loop handed `Monitor` a shell `while true; … sleep 300; done`. On macOS that eval context strips `PATH` (so `sleep`/`date` resolve to "command not found") and kills backgrounded children with `nice(5) failed: operation not permitted`. New `watch.ts` runs one `bun` process that sleeps internally via `setTimeout`, matching the pattern already used by `github:actions-monitor`'s `watch.ts`. It composes the existing `state.ts sync` and `poll.ts` fetch-and-diff logic rather than duplicating them; `poll.ts` keeps its one-shot mode for the interactive flow. It carries the `claude:dangerouslyDisableSandbox` marker, since it shells out to `gh`/`wt`/`tmux` for TLS-bearing calls under `Monitor` (without it, the queue commands hit the exact TLS failure this PR addresses).

#### Guidance

Dashboard `SKILL.md` swaps the shell-loop snippet for the `watch.ts` invocation and adds a short Monitor Reliability note: hand `Monitor` a single long-lived process, never suppress poll output, and prefer REST (`gh api repos/<o>/<r>/commits/<sha>/check-runs`) over GraphQL (`gh pr view --json statusCheckRollup`) for CI/review state under the sandbox, where GraphQL TLS verification fails intermittently. `babysit` and `follow-up` cross-reference it.

## Testing

Added `fetchUrls` tests covering each branch of the new result: success with URLs, empty queue, non-zero exit with and without stderr, and unparseable JSON. The existing `newUrls` tests are unchanged. Also smoke-tested `watch.ts` end to end with one healthy and one failing queue, confirming the new URL prints to stdout while the failing source emits a `source-error` line to stderr.

Original Task: [Fix: babysit/Monitor poll silently fails on sandbox TLS errors (gh GraphQL)](https://things.bendrucker.me/show?id=RCzjGWHG9R4HpfraYDnaVo)
Original Task: [Claude Code: Monitor tool broken on macOS](https://things.bendrucker.me/show?id=Bevw45QXq4o9fYwFG5cKV2)

### DIFF
--- a/plugins/pull-request/skills/babysit/SKILL.md
+++ b/plugins/pull-request/skills/babysit/SKILL.md
@@ -132,3 +132,5 @@ The Bash tool escapes `!` to `\!`. Use `| not` in jq filters (e.g. `select(.x ==
 The watcher dedupes by `(sha, state)`. A `failing` event for a SHA older than `git rev-parse HEAD` means a fix was already pushed; ignore it.
 
 Babysit is session-scoped. If the session ends, the watcher process ends with it. Re-invoke this skill from a new session to resume monitoring.
+
+Whatever you hand `Monitor`, follow the Monitor Reliability rules in `review:dashboard`: a single long-lived process that sleeps internally (never a shell `while/sleep` loop, which breaks on macOS); never suppress poll output, so a silent poll is never mistaken for "nothing new"; prefer REST (`gh api repos/<o>/<r>/commits/<sha>/check-runs`) over GraphQL (`gh pr view --json statusCheckRollup`) for CI state under the sandbox, where GraphQL TLS verification fails intermittently.
--- a/plugins/pull-request/skills/babysit/SKILL.md
+++ b/plugins/pull-request/skills/babysit/SKILL.md
@@ -132,5 +132,3 @@ The Bash tool escapes `!` to `\!`. Use `| not` in jq filters (e.g. `select(.x ==
 The watcher dedupes by `(sha, state)`. A `failing` event for a SHA older than `git rev-parse HEAD` means a fix was already pushed; ignore it.
 
 Babysit is session-scoped. If the session ends, the watcher process ends with it. Re-invoke this skill from a new session to resume monitoring.
-
-Whatever you hand `Monitor`, follow the Monitor Reliability rules in `review:dashboard`: a single long-lived process that sleeps internally (never a shell `while/sleep` loop, which breaks on macOS); never suppress poll output, so a silent poll is never mistaken for "nothing new"; prefer REST (`gh api repos/<o>/<r>/commits/<sha>/check-runs`) over GraphQL (`gh pr view --json statusCheckRollup`) for CI state under the sandbox, where GraphQL TLS verification fails intermittently.
--- a/plugins/pull-request/skills/follow-up/SKILL.md
+++ b/plugins/pull-request/skills/follow-up/SKILL.md
@@ -67,3 +67,5 @@ The gated default requires checking before any post or resolve. `--auto` lifts t
 - Resolve only after replying; silent resolves hide context.
 - Never name or thank the bot in a reply; write it as a note for any reader. The `@<bot>` re-trigger is the one exception.
 - Match my writing style; you're replying as me.
+
+When polling review or CI state under the sandbox, prefer REST (`gh api repos/<o>/<r>/commits/<sha>/check-runs`) over GraphQL (`gh pr view --json statusCheckRollup`): GraphQL TLS verification fails intermittently there. See the Monitor Reliability rules in `review:dashboard`.
--- a/plugins/pull-request/skills/follow-up/SKILL.md
+++ b/plugins/pull-request/skills/follow-up/SKILL.md
@@ -67,5 +67,3 @@ The gated default requires checking before any post or resolve. `--auto` lifts t
 - Resolve only after replying; silent resolves hide context.
 - Never name or thank the bot in a reply; write it as a note for any reader. The `@<bot>` re-trigger is the one exception.
 - Match my writing style; you're replying as me.
-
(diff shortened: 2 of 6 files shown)


plugins/pull-request/skills/babysit/SKILL.md
```diff
@@ -132,3 +132,5 @@ The Bash tool escapes `!` to `\!`. Use `| not` in jq filters (e.g. `select(.x ==
 The watcher dedupes by `(sha, state)`. A `failing` event for a SHA older than `git rev-parse HEAD` means a fix was already pushed; ignore it.
 
 Babysit is session-scoped. If the session ends, the watcher process ends with it. Re-invoke this skill from a new session to resume monitoring.
+
+Whatever you hand `Monitor`, follow the Monitor Reliability rules in `review:dashboard`: a single long-lived process that sleeps internally (never a shell `while/sleep` loop, which breaks on macOS); never suppress poll output, so a silent poll is never mistaken for "nothing new"; prefer REST (`gh api repos/<o>/<r>/commits/<sha>/check-runs`) over GraphQL (`gh pr view --json statusCheckRollup`) for CI state under the sandbox, where GraphQL TLS verification fails intermittently.
```
bendrucker: No cross-plugin references.

plugins/pull-request/skills/follow-up/SKILL.md
```diff
@@ -67,3 +67,5 @@ The gated default requires checking before any post or resolve. `--auto` lifts t
 - Resolve only after replying; silent resolves hide context.
 - Never name or thank the bot in a reply; write it as a note for any reader. The `@<bot>` re-trigger is the one exception.
 - Match my writing style; you're replying as me.
+
+When polling review or CI state under the sandbox, prefer REST (`gh api repos/<o>/<r>/commits/<sha>/check-runs`) over GraphQL (`gh pr view --json statusCheckRollup`): GraphQL TLS verification fails intermittently there. See the Monitor Reliability rules in `review:dashboard`.
```
bendrucker: Same thing, no cross-plugin refs.

plugins/review/skills/dashboard/SKILL.md
```diff
@@ -129,21 +129,27 @@ A platform plugin that owns its queue keeps the query; the dashboard only runs t
...
 ```
 
-`sync`'s output is suppressed so that only new-review URLs become events. Its worktree-removal failures flow to the monitor's output file, readable with `Read` if a prune fails. `poll.ts` treats a failed source as an empty queue, so one source's outage never stalls the loop.
+A failed source emits a `{"type":"source-error",...}` line to stderr and contributes zero URLs, so one source's outage never stalls the loop. Sync errors go to stderr as `{"type":"sync-error",...}`. Neither is suppressed.
+
+#### Monitor Reliability
+
+Three rules that apply here and anywhere you hand a shell command to `Monitor`:
+
+- Pass a **single long-lived process** that sleeps internally via `setTimeout`. Never a shell `while/sleep` loop: Monitor's eval context strips `PATH`, so `sleep` and `date` are not found, and backgrounded children are killed by `nice(5) failed: operation not permitted`.
```
bendrucker: This one is kind of crazy and hopefully temporary. We need some way to do shell monitors...

plugins/review/skills/dashboard/SKILL.md
```diff
@@ -129,21 +129,27 @@ A platform plugin that owns its queue keeps the query; the dashboard only runs t
...
+- **Never suppress poll output** with `>/dev/null` or `|| true`. If a poll is silent, it is indistinguishable from "nothing new." Emit a structured line on error so failures are visible.
+- Prefer **REST over GraphQL** for CI/review state under the Claude Code sandbox (`gh api repos/<o>/<r>/commits/<sha>/check-runs` works; `gh pr view --json statusCheckRollup` fails intermittently with TLS certificate errors).
```
bendrucker: This feels like a coincidence that Claude misinterpreted their TLS errors because of sandbox issues running `gh` (all Go CLIs), idk what that has to do with either command.

plugins/pull-request/skills/babysit/SKILL.md
bendrucker: Removed in 17698b71. The note crossed plugin boundaries, and babysit hands CI watching to the provider skill rather than building Monitor commands itself.

plugins/pull-request/skills/follow-up/SKILL.md
bendrucker: Removed in 17698b71, along with the cross-plugin reference.

plugins/review/skills/dashboard/SKILL.md
bendrucker: Reframed in 17698b71 as a temporary workaround for the macOS Monitor eval-context bug, not a permanent rule. Shell loops should work again once the harness bug is fixed.

plugins/review/skills/dashboard/SKILL.md
bendrucker: Removed in 17698b71. Agreed it was a misdiagnosis: the TLS failure (OSStatus -26276) is gh's Go TLS stack failing through the sandbox proxy, not GraphQL- or endpoint-specific (gh pr checks uses GraphQL too). The real mitigation is the dangerouslyDisableSandbox marker watch.ts already carries.

PR #735, opened 2026-06-05
writing: exempt plan files from the numbering hook

bendrucker:
The numbering hook fires on numbered markdown headings and numbered code identifiers (`step1`, `phaseN`). Across the full session history every block came from markdown headings, and 82% of those were plan-mode files in `~/.claude/plans/`, where an ordered list of steps is expected. The code-identifier path has never fired.

Gating the hook to markdown (the original approach) would have left every plan-file block in place and disabled the never-firing code path, so it reduced no friction and dropped coverage.

Instead, exempt plan files the way memory files are already exempted. Deliverables staged in project `tmp/` directories, like PR bodies passed to `gh`, still get validated on write.

#### Changes

- `isPlanPath()` in `markdown.ts`, matching `~/.claude/plans/`
- `numbering.ts` skips plan files alongside memory files
- Reverts the markdown gate on the numbering hook; `headings.ts` stays markdown-gated
- Tests covering plan-file exemption and a numbered heading in `tmp/` still denying

### DIFF
--- a/plugins/writing/hooks/markdown.ts
+++ b/plugins/writing/hooks/markdown.ts
@@ -17,6 +17,12 @@ export function isMemoryPath(filePath: string): boolean {
   return MEMORY_PATH_PATTERN.test(resolve(filePath));
 }
 
+const PLAN_PATH_PATTERN = /\/\.claude\/plans\//;
+
+export function isPlanPath(filePath: string): boolean {
+  return PLAN_PATH_PATTERN.test(resolve(filePath));
+}
+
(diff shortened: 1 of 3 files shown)


bendrucker:
The heading hook only is meant to apply to Markdown, but the hook for numbering is meant to apply to code files as well because Claude often writes crappy code comments in there where it creates things like numbered phases unnecessarily.

bendrucker:
Superseded by #739, which already exempts plan files (and plan mode) from the numbering hook on main. Closing as a duplicate.

PR #736, opened 2026-06-05
pull-request: bound babysit self-pacing to cap runaway token spend

bendrucker:
## What changed and why

The `pull-request:babysit` skill drives a hands-off CI watch and, under `/loop` self-pacing, had no self-imposed ceiling on how long or how many cycles it would run. One unattended babysit/loop session burned 15.4M output tokens (16% of a 95.8M-token corpus) over 24.7 hours, with 189 turns hitting the 32k+ output cap (local). This adds an explicit `## Bounds` section so a single babysit invocation stays bounded, and tightens the frontmatter description to one trigger sentence.

## Evidence

- 15.4M output tokens in a single session, 16% of the 95.8M-token corpus (local).
- 24.7 hours wall-clock for that one session (local).
- 189 turns hitting the 32k+ output cap (local).

## The change

`plugins/pull-request/skills/babysit/SKILL.md` is the only file touched. It is prose-only; no tool list, script, or event schema changes.

- Wall-clock ceiling: babysit honors the watcher's existing `--max-minutes` (default 60) via the `max-time-reached` event it already handles. Babysit has no timer of its own, so the only wall clock it can enforce is the one the watcher emits. Reusing it avoids inventing a second clock the skill cannot honor.
- No silent re-arm: on `max-time-reached`, babysit reports cumulative work and stops rather than starting a fresh watcher generation. Under `/loop`, the loop owns repetition. The 16% burn came from indefinite looping, so making each invocation single-generation is the decisive cut.
- Iteration ceiling for the `--merge` poll: that loop is babysit's only self-driven poll and previously had only a 3-submit oscillation guard, no wall-clock or count bound. It now stops after 20 polls or the 60-minute wall clock, whichever comes first. The existing 3-submit guard is retained alongside it ("whichever first") so neither weakens the other.
- Long intervals and prompt cache: CI and merge state change on the order of minutes, so polls run at 60s or longer (preferring the watcher's derived or 180s interval) and never tight-loop. This targets the 189 turns at the output cap and the prompt-cache churn.
- Cheapest viable model for poll/wait/status-read steps, matching the `github:logs` agent that already runs on haiku; fixes and merge decisions stay on the default model.

All values are defaults the user can override via `$ARGUMENTS`.

The frontmatter description drops from 357 chars to 135 (one trigger sentence plus the two flags). This file is excluded from the in-flight skill-description PR, so the description edit will not conflict.

## Verification

- `bun run skill-lint "plugins/pull-request/skills/babysit"`: 0 errors, 0 warnings; `description-length` now 135 (was 357), `body-lines` 143 (max 500).
- `bun scripts/check-marketplace.ts`: unaffected, confirmed nothing structural broke.
- No tests added or changed: babysit has no test file, and the watcher tests stay untouched because no flag or script behavior changed.

### DIFF
--- a/plugins/pull-request/skills/babysit/SKILL.md
+++ b/plugins/pull-request/skills/babysit/SKILL.md
@@ -1,7 +1,6 @@
 ---
 name: pull-request:babysit
-description: |
-  Monitor a PR's CI, fix trivial failures (lint, types, formatting), and self-cancel when green. Use after pushing for hands-off CI monitoring. With --merge, drive the PR to merged (handling merge-train/queue kickouts, rebase issues, and re-runs); with --reviews, hand off to AI-review triage after the first green.
+description: Monitor a PR's CI, fix trivial failures, and self-cancel when green; --merge drives to merged, --reviews hands off to AI-review triage.
 argument-hint: "[--merge] [--reviews]"
 allowed-tools:
   - Monitor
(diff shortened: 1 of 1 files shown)


bendrucker:
Headings need to be title cased, among other sloppy formatting issues

PR #750, opened 2026-06-07
tmux: remove team-workaround hook

bendrucker:
Removes the tmux `team-workaround` hook entirely: the script, its tests, the `shell-wrapper.sh` it installs, both `hooks.json` entries (the `PreToolUse` `TeamCreate` setup matcher and the Stop teardown), and the README line.

The teardown fired on every Stop (1,089 fires across 166 sessions since 2026-05-07) while the paired setup never fired locally, so every run was a no-op bun spawn. The upstream race it works around ([anthropics/claude-code#23513](https://github.com/anthropics/claude-code/issues/23513)) is closed "not planned", but agent teams see little to no use here, so I deleted the workaround rather than gating or relocating it.

## Testing

- The remaining tmux plugin tests (`target.test.ts`) still pass after the deletion
- `jq` confirms `hooks.json` is valid with only the surviving `SessionStart`/`PreToolUse`/`Notification`/`Stop` entries
- `grep` finds no remaining references to `team-workaround` or `shell-wrapper`

Original Task: [[discovery] team-workaround rip-out](https://things.bendrucker.me/show?id=KVqqt73onAxsRkgUwnAHfo)


plugins/tmux/hooks/team-workaround.ts
```diff
@@ -1,79 +0,0 @@
...
-//
-// Agent team tmux panes fail because send-keys fires before zsh finishes
-// loading .zshrc. We override default-command with a wrapper script that
-// checks a per-window option (@claude_fast_shell). Windows with the option
-// get instant bash; all other windows fall through to a normal login shell.
-
-import { join } from "node:path";
-import { cli } from "cleye";
-import { tmux, tmuxQuery } from "./tmux";
-
```
bendrucker, line 13: Deleting this file removes the only consumer of the `tmux()` export in `plugins/tmux/hooks/tmux.ts` — `context.ts` only imports `tmuxQuery`, and `notification.ts` defines its own local `tmux()`. The export is now dead code; consider removing it (or having `notification.ts` use it instead of its local copy).

bendrucker: Fixed in b26b05b7: notification.ts now imports the shared tmux() and tmuxQuery() from ./tmux instead of its local duplicates, so the export has a consumer again.

PR #759, opened 2026-06-07
session: dedup tokens, fix retry and repeat-read queries

bendrucker:
Fixes four analysis bugs in the `claude-code:session` skill surfaced by the W23 discovery run, and trims the SKILL.md frontmatter description.

## Changes

- Adds a `message_usage` view that dedupes assistant usage to one row per message id (`MAX` per token column). Claude Code writes one JSONL row per content block and each row repeats the parent message's cumulative usage, so per-row sums inflated output totals 2-3.5x. `top-sessions-by-output` and `skill-activity` now read from the view, and SKILL.md/CLAUDE.md document it as the token-summing surface.
- Rewrites `hook-block-then-retry-success` NULL-safe. `NOT s.blocked` excluded nearly every success row (the view's `blocked` boolean is NULL when `decision` is NULL), so every hook structurally reported 0% retry. Successes are now `kind = 'hook_success' AND blocked IS NOT TRUE`, and Stop rows key by `hook_event` because blocked Stop events carry NULL `command` while successes carry it, so a command-based key never aligned the two sides.
- Decomposes `repeat-read-waste` by cause: paginated (`offset`/`limit` chunking), sidechain (subagent fan-out), after-own-edit, and true repeats. The old headline percentage mostly measured pagination and fan-out architecture, not waste.
- Excludes `ExitPlanMode` and `AskUserQuestion` from the `permissions` query and from rejection rows in `errors`: blocking on user input is those tools working, not friction a setting can remove. `references/discovery.md` now points analysts at the structural exclusion instead of manual filtering.
- Documents that `SET VARIABLE limit = 5` is a parser error (`limit` is a reserved word) and must be quoted as `"limit"`, and compresses the frontmatter description from 385 to 312 chars with the activation-trigger phrases intact.

## Testing

Validated against the live index (`refresh.ts`, then `duckdb -readonly`):

- Output-token inflation collapses from 3.13x (local) under the old per-row sum to deduped totals; the previous #1 "runaway" session reads 504,217 output tokens deduped vs 15,455,129 naive, confirming the 30x artifact from the digest.
- The retry-success query returns non-zero rates (e.g. 31/37 local Stop blocks retried away within the window) where it structurally returned 0% before.
- Repeat-read buckets on the all-time corpus: 5,751 paginated / 7,320 sidechain / 951 main-thread unpaginated across hosts, matching the digest grounding (5,712 / 7,275 / 882) plus corpus growth since.
- `bun run skill-lint` passes on the session skill.

Original Task: [[discovery] token dedup](https://things.bendrucker.me/show?id=PXbfnShpk1zoDRDvVpCu8n)
Original Task: [[discovery] retry-success query](https://things.bendrucker.me/show?id=PaRMKTQsQgxv5wKSXPmdu2)
Original Task: [[discovery] repeat-read query](https://things.bendrucker.me/show?id=NKphCembZRmAd2txtFC5Zx)
Original Task: [[discovery] permissions query input-tool exclusion](https://things.bendrucker.me/show?id=y8WLZTPeVk757zJoLnszr)
Original Task: [[discovery] description trims](https://things.bendrucker.me/show?id=4H8HoLXZEQtTFJHYuAPrWK)


plugins/claude-code/skills/session/resources/queries/repeat-read-waste.sql
```diff
@@ -1,44 +1,69 @@
...
   COUNT(*)                                       AS total_reads,
   COUNT(*) FILTER (WHERE rn > 1)                 AS repeat_reads,
   ROUND(100.0 * COUNT(*) FILTER (WHERE rn > 1) / COUNT(*), 1) AS repeat_pct,
-  SUM(chars) FILTER (WHERE rn > 1)               AS repeat_chars,
+  COUNT(*) FILTER (WHERE rn > 1 AND paginated)    AS paginated,
+  COUNT(*) FILTER (WHERE rn > 1 AND is_sidechain) AS sidechain,
+  COUNT(*) FILTER (WHERE rn > 1 AND NOT paginated AND NOT is_sidechain AND after_own_edit)
+    AS after_own_edit,
+  COUNT(*) FILTER (WHERE rn > 1 AND NOT paginated AND NOT is_sidechain AND NOT after_own_edit)
+    AS true_repeats,
```
bendrucker, line 83: `true_repeats` counts main-thread reads whose only prior read happened in a sidechain. `rn` partitions by `(host, session_id, file_path)`, and subagent files carry the parent session's id, so a subagent's read takes `rn=1`; when the main thread then reads the file for the first time in its own context, that row has `rn=2`, `is_sidechain=false`, and (absent pagination/edits) lands in `true_repeats` — but nothing was re-injected into any context. With heavy fan-out this inflates exactly the bucket the split exists to isolate.

A fix is to require a prior *main-thread* read before classifying a repeat as a true repeat, e.g. a windowed count of prior non-sidechain reads in `ranked`:

```sql
COUNT(*) FILTER (WHERE NOT is_sidechain) OVER (
  PARTITION BY host, session_id, file_path
  ORDER BY timestamp, source_line
  ROWS BETWEEN UNBOUNDED PRECEDING AND 1 PRECEDING
) AS prior_main_reads
```

and gate `true_repeats`/`after_own_edit` on `prior_main_reads >= 1` instead of `rn > 1`.

plugins/claude-code/skills/session/resources/queries/repeat-read-waste.sql
```diff
@@ -1,44 +1,69 @@
...
-    ) AS rn
-  FROM reads
+    ) AS rn,
+    EXISTS (
+      SELECT 1 FROM tool_calls e
+      WHERE e.host = r.host
+        AND e.session_id = r.session_id
+        AND e.file_path = r.file_path
+        AND e.tool_name IN ('Write', 'Edit', 'MultiEdit', 'NotebookEdit')
+        AND e.timestamp <= r.timestamp
```
bendrucker, line 68: `e.timestamp <= r.timestamp` matches any edit ever before this read, so a single edit early in a session permanently classifies every later repeat read of that file as `after_own_edit` — even when no edit happened between consecutive reads. A session that edits a file once and then re-reads it five times reports zero `true_repeats`, though four of those reads are exactly the context tax the query is meant to surface (the header's "refreshing post-edit state" only holds when an edit intervenes between reads).

Bounding the window to edits since the previous read fixes it: compute `lag(timestamp) OVER (PARTITION BY host, session_id, file_path ORDER BY timestamp, source_line)` in `ranked` and add `AND e.timestamp > prev_read_ts` to the EXISTS.

plugins/claude-code/skills/session/resources/queries/repeat-read-waste.sql
bendrucker: Fixed in de63b54: `ranked` now computes `prior_main_reads` (windowed count of prior non-sidechain reads, unbounded preceding to 1 preceding) and the `after_own_edit`/`true_repeats` buckets gate on `prior_main_reads >= 1` instead of `rn > 1`. A main-thread read whose only prior reads were sidechain no longer lands in either actionable bucket. Headline `repeat_reads`/`repeat_pct` still use `rn > 1`.

plugins/claude-code/skills/session/resources/queries/repeat-read-waste.sql
bendrucker: Fixed in de63b54: the EXISTS now also requires `e.timestamp > prev_main_read_ts`. Instead of plain `lag()`, the bound is a windowed `MAX(timestamp) FILTER (WHERE NOT is_sidechain)` over prior rows, so the edit window is measured against the last main-thread read (the last time the file was actually in main context) rather than a possibly-sidechain previous row. Verified the read → edit → 3x read case now yields one `after_own_edit` and two `true_repeats`.

PR #792, opened 2026-06-10
writing: encode detector layers, lifecycle metadata, and pattern examples

### LINKED ISSUE #787: writing: encode detector layers and rule lifecycle in the engine
The four-layer detector model (vocabulary / grammar / cross-sentence / meaning) and the curation principle both live in prose (`linguistics.md`, CLAUDE.md). The engine does not represent either, which produced concrete drift found in the 2026-06-09 codebase audit. This issue makes the model and the lifecycle structural so every future detector lands with them.

## Layer Field on PatternDef

Add `layer: "vocabulary" | "grammar" | "cross-sentence" | "meaning"` to `PatternDef`. Today `CROSS_SENTENCE_NOT` and `connectorDensityHits` are cross-sentence detectors sitting unlabeled in the flat `PATTERNS` array in `detection/tropes.ts`, shipped to the hook without ever passing the linguistics promotion bar. The `structural.ts` audit already labels patterns by scope and should also report by layer, which makes misplaced detectors visible.

## Embedded Examples per Pattern

Adopt the convention LanguageTool (`<example>` elements doubling as unit tests) and textlint (`valid`/`invalid` fixture tables) both converged on: each `PatternDef` carries committed `positives: string[]` and `negatives: string[]` arrays that drive a deterministic regression test. This generalizes the `linguistics/fixtures.ts` pattern from headings to every rule and is the committable fixture suite #769 calls for. Examples are invented, never quoted from sessions.

## Retirement Metadata

Wordlist entries have an audit-driven retirement path (dead / not distinctive). The 22 regex and function patterns have none, which violates the curation principle. Each pattern should record what evidence earned it and what corpus condition removes it, either as a structured field or a required comment block, and the structural audit should surface patterns whose removal condition is met.

## Meaning-Layer Seam

No code location exists where an LLM-judge batch pass would plug in. Document `structural.ts` as the template for batch detector modules (rows in, typed findings out, wired in `runAnalysis`, rendered via `ReportInput`) and define the meaning-layer detector interface against it, even before any judge ships.

## Consolidations

- `countSubsequence` is duplicated between `detection/wordlists.ts` and `skills/analyze/scripts/deliverable-audit.ts`. A divergence would make the hook and the audit count the same phrase differently. Export one copy.
- `readInput` is duplicated between `skills/scan/scripts/scan.ts` and `skills/score/scripts/score.ts`.
- `stripCode` (offset-preserving, `detection/tropes.ts`) and `cleanText` (aggressive, `skills/analyze/scripts/ngram.ts`) are intentionally different cleaning pipelines with no documented contract. Document the split or converge on a shared base strip.
- `skills/review/SKILL.md` injects eight vocabulary words (`pivotal`, `testament`, `underscore`, `interplay`, `intricacies`, `bolstered`, `garner`, `foster`) that appear in no wordlist, so review flags what the hook passes. Either promote them through the corpus audit into `vocabulary.txt` or document why they are review-only.

Refs #769.

bendrucker:
Adds required `layer`, `evidence`, `retire`, `positives`, and `negatives` fields to every `PatternDef` and `WeightedPatternGroup`. Introduces `examples.test.ts` as a deterministic regression gate for each pattern's examples. Closes #787.

## Changes

- `PatternDef` gains `layer: DetectorLayer`, `evidence`, `retire`, `positives[]`, and `negatives[]`. Every pattern in `PATTERNS` is labeled. `WeightedPatternGroup` gains `layer`, `evidence`, and `retire`.
- `examples.test.ts` drives a regression test over all embedded examples: each positive must match, each negative must not, and every pattern must have at least 2 of each.
- `structural.ts` audit now groups patterns by layer in the report and prints the `retire` condition alongside hit counts. `StructuralAuditRow` gains `layer` and `retire`.
- `DetectorModule<TFinding>` interface added to `structural.ts` as the seam for future batch detector modules (meaning-layer judges, grammar-tagger detectors). `detect()` returns `Promise<TFinding[]>`.
- `countSubsequence` exported from `detection/wordlists.ts`; the duplicate in `deliverable-audit.ts` removed.
- `readInput` extracted to `skills/io.ts`; both `scan.ts` and `score.ts` import from there. `scan.ts` re-exports it to keep its test import unchanged.
- `stripCode` vs `cleanText` contracts documented at their definitions and in `methodology.md`. The two pipelines are intentionally different.
- `skills/review/SKILL.md`: the eight review-only words documented with the per-surface rationale (recall-leaning review vs precision-critical hook; promotion requires a corpus audit).

## Testing

`examples.test.ts` gates every pattern's positives (must match), negatives (must not match), layer label, evidence string, and retirement condition. The existing analyze test suite covers `StructuralAuditRow` shape, report rendering, and the shared `countSubsequence` path through `deliverable-audit`.


plugins/writing/skills/review/SKILL.md
```diff
@@ -31,6 +31,7 @@ Spawn all applicable agents in parallel. Each receives the document content, aud
 For `writing:style`, also inject these writing preferences (sub-agents cannot see skills):
 
 - Avoid AI-typical vocabulary: `meticulous`/`meticulously`, `pivotal`, `testament`, `underscore` (figurative), `interplay`, `intricacies`, `bolstered`, `garner`/`garnered`, `foster`/`fostering`
+  - These eight words are **review-only**. They do not appear in the hook wordlists because none has a corpus audit behind it. The review skill runs in batch mode where a human triages flags (recall matters more than precision), matching the analyze-and-review bar from linguistics.md. Promoting any of them to the hook wordlists requires a corpus audit confirming lift and distinctiveness versus the user's voice baseline.
```
bendrucker: "eight words" useless counting

plugins/writing/scripts/io.ts
bendrucker, line 1: This is an improper location for a shared script. It should be at the plug‑in level scripts directory if multiple skills need to share it.

PR #804, opened 2026-06-12
user: add job skill for daily work routines

bendrucker:
Adds a user-level `/job` skill encoding day-job routines: `/job start` triages the inbound review queue, your own open PRs, and the tracker's plan for today. `/job end` clears the outbox, surfaces review debt, sweeps worktrees for unpushed work, and tidies tracker state for tomorrow. Both modes gather read-only through parallel sub-agents, present one prioritized brief, then execute confirmed safe actions. Approve, close, and merge stay ask-first.

The skill is publishable: it names no platforms, hosts, or usernames. A first-run `/job setup` interview detects installed skills and CLIs, asks which tools apply, and persists the answers to `~/.config/claude-job-skill/config.json`. Each run bang-injects that config and delegates platform mechanics to whatever installed skill matches the configured tool, falling back to the configured CLI. I put the config under `~/.config` rather than `~/.claude` because `user/` symlinks into this public repo.

`disable-model-invocation: true` keeps recurring context cost at zero. The curation signal is `/job` usage in session history. Removal is deleting the directory.

## Testing

Verified with strict `skill-lint` (frontmatter, references, bang syntax) and a `writing:scan` pass over the prose, plus a grep of all four files for platform names, hostnames, and usernames to confirm nothing employer-specific leaked in.

Not exercised end-to-end. This was written on my home machine, so the live `/job setup` and mode runs happen on the work machine after merge, with revisions to follow.

### DIFF
--- a/user/skills/job/references/end.md
+++ b/user/skills/job/references/end.md
@@ -16,8 +16,9 @@ Dispatch parallel read-only sub-agents:
 Everything pushed today should leave tonight with:
 
 - A reviewer assigned. Assigning one is safe; respect any reviewer conventions in the config notes.
-- A self-review pass before anything new goes out for review. Delegate to an installed code-review skill.
 - A body that matches the current diff. When commits have outrun the description, delegate to an installed PR-update skill.
+
+A PR/MR sitting without a reviewer is usually deliberate: doubt about the approach, a blocker, a dependency. Surface it with the likely reason and ask for the next step rather than defaulting to another review pass. Offer a self-review via an installed code-review skill only when the user wants one before sending.
 
 Flag items that need collaboration lead time (a reviewer in another timezone, a long CI run) so they go out tonight rather than tomorrow morning.
 
@@ -31,6 +32,8 @@ Review debt is inbound requests you did not reach today plus unanswered threads 
 Never silently drop an item.
 
 ## Unpushed Work Sweep
+
+The configured roots may span many repositories, so keep this phase cheap: one sub-agent, local git commands only, no platform API calls. This is the lowest-priority phase. Skip it when the focus hint points elsewhere or the rest of the brief already needs the time.
 
 Enumerate worktrees via the configured worktree tool. Fall back to `git worktree list` run across the configured roots. For each worktree, report:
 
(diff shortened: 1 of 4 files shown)


user/skills/job/references/end.md
```diff
@@ -0,0 +1,49 @@
...
+- Inbound: review requests still open against you, and threads on others' PRs/MRs awaiting your reply.
+- Worktrees: every worktree's uncommitted and unpushed state (details under the sweep below).
+- Tracker: your in-flight issues and their current states.
+
+## Outbox Clearing
+
+Everything pushed today should leave tonight with:
+
+- A reviewer assigned. Assigning one is safe; respect any reviewer conventions in the config notes.
+- A self-review pass before anything new goes out for review. Delegate to an installed code-review skill.
```
bendrucker: This could be duplicative because I'm normally doing this myself anyway. This is really only an issue for something that I've created but not assigned a reviewer for immediately. That's typically because there is some doubt about the approach, or it's blocked on something, or similar. I don't always just want to do more automated code reviews. It's typically a sign that I need to weigh in with some sort of next step.

user/skills/job/references/end.md
```diff
@@ -0,0 +1,49 @@
...
+## Review Debt
+
+Review debt is inbound requests you did not reach today plus unanswered threads on others' PRs/MRs. Choose one path per item:
+
+- Reply now: draft the reply and include it in the brief. Posting a draft the user approved is safe.
+- Carry to tomorrow: record it explicitly so it appears in tomorrow's start brief.
+
+Never silently drop an item.
+
+## Unpushed Work Sweep
```
bendrucker, line 34: This part doesn't acknowledge multiple repositories. This would have to fan out across repositories, which might prove expensive. If this is a major distraction for now, this is not an especially important part of the workflow.

user/skills/job/references/end.md
bendrucker: Dropped the default self-review pass in 862a7965. A reviewer-less PR now surfaces with its likely reason (doubt about approach, blocker) and asks you for the next step. Code review is offered only when you want one before sending.

user/skills/job/references/end.md
bendrucker: Acknowledged multi-repo roots in 862a7965 and bounded the cost: one sub-agent, local git commands only, no API calls. The sweep is now the lowest-priority phase and gets skipped when focus is elsewhere.

PR #857, opened 2026-06-19
improve-claude-code: build weekly Discover cadence

bendrucker:
Builds the weekly Discover cadence the skill already documented but never stood up. Adds a non-interactive run path so a scheduled local run produces the digest and one Things doorway item, instead of blocking on the interactive "which to file" prompt. Discover still never auto-files keepers. The doorway hands triage back to you on demand, and the mandatory grounding pass stays exactly as it is on-demand.

The trigger must be local. The session DB and `duckdb` live on this Mac, and cloud `/schedule` routines run from a fresh clone with no local access. The mechanism is a Claude Code Desktop scheduled task (Routines page, Local mode) running `/improve-claude-code discover` weekly. A launchd job invoking `claude -p` is the documented fallback for when the app is closed, and that plist is the paired dotfiles change.

An earlier draft justified the Desktop task with a billing claim: that headless `claude -p` draws from a separate metered credit pool. That is false, verified against the Claude Code cost and headless docs. Both bill against the same subscription. The claim is removed.

## Testing
- `bun run skill-lint "user/skills/improve-claude-code"`
- A paired dotfiles PR (the launchd fallback) is still needed and is not built here.


bendrucker:
Closing as superseded by #974 (scheduled discover mode with launchd agent, merged 2026-07-07), which solved the same weekly-cadence problem with the opposite design: it auto-files high-confidence candidates and commits the launchd plist in-repo, whereas this PR deliberately never auto-files (one Things doorway item) and keeps the plist in dotfiles. A faithful rebase onto current main isn't possible without leaving SKILL.md self-contradictory. If the doorway-item model or dotfiles-owned plist is still wanted, worth reproposing against the current `discover --scheduled` design.

PR #871, opened 2026-06-22
review: dispatch reviews to a named tmux session

bendrucker:
Adds a `--session <name>` flag to `spawn.ts` so reviews can open in a tmux session other than the one running the inbox.

`spawn.ts` split `$TMUX_PANE`, so panes landed wherever the orchestrator ran. That is fine when you run the inbox as a sidebar inside your review session. It breaks once a remote orchestrator dispatches the queue: a `job` start run in your work session was splitting its own pane, so the review panes leaked into that session instead of your dedicated review session.

With `--session reviews`, `layout.ts` targets that session's active pane for the first review and skips the sidebar carve-out, since the orchestrator does not live there. Later panes still chain off the last review pane, which now sits in the target session, so the group-of-three column cycling is unchanged. Omit the flag and behavior is identical to before. The target path also drops the `TMUX_PANE` requirement, since a remote orchestrator need not be in tmux at all.

The layout tests gain cases for the target-session path: the first pane splits the named session, that path no longer needs `TMUX_PANE`, and later panes still chain off the last review pane.

For the actual `job → inbox` handoff to use this, the curated dispatch (#867, #868) has to pass `--session`, with the session name read from `job` config. That wiring is a follow-up on those branches.

### DIFF
--- a/plugins/review/skills/inbox/SKILL.md
+++ b/plugins/review/skills/inbox/SKILL.md
@@ -79,7 +79,7 @@ bun ${CLAUDE_SKILL_DIR}/scripts/spawn.ts <pr-url> --repo-path <local-path> --dat
 
 `spawn.ts` handles `--worktree` for branch isolation, tmux layout computation, and state tracking. Pass `--context` with PR metadata (title, author, description summary) so the spawned review session has immediate context. Panes cycle in groups of 3: one horizontal split (new column at 70% width for the first, equal width after), then two vertical splits stacking in the column.
 
-Pass `--session <name>` to open reviews in a tmux session other than the one you orchestrate from. Without it, `spawn.ts` splits the orchestrator's own pane. That fits running the inbox as a sidebar inside the review session itself. With it, a remote orchestrator dispatches elsewhere: a `job` start run in your work session hands the queue here, and the reviews open in your dedicated review session. The first pane splits that session's active pane rather than carving a sidebar from the orchestrator.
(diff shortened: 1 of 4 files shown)


plugins/review/skills/inbox/SKILL.md
```diff
@@ -79,7 +79,9 @@ bun ${CLAUDE_SKILL_DIR}/scripts/spawn.ts <pr-url> --repo-path <local-path> --dat
 
 `spawn.ts` handles `--worktree` for branch isolation, tmux layout computation, and state tracking. Pass `--context` with PR metadata (title, author, description summary) so the spawned review session has immediate context. Panes cycle in groups of 3: one horizontal split (new column at 70% width for the first, equal width after), then two vertical splits stacking in the column.
 
-Before spawning the first pane, resize the orchestrator to a sidebar:
+Pass `--session <name>` to open reviews in a tmux session other than the one you orchestrate from. Without it, `spawn.ts` splits the orchestrator's own pane. That fits running the inbox as a sidebar inside the review session itself. With it, a remote orchestrator dispatches elsewhere: a `job` start run in your work session hands the queue here, and the reviews open in your dedicated review session. The first pane splits that session's active pane rather than carving a sidebar from the orchestrator.
```
bendrucker:
```suggestion
Pass `--session <name>` to open reviews in a tmux session other than the one you orchestrate from. Without it, `spawn.ts` splits the orchestrator's own pane.
```

PR #879, opened 2026-06-24
session: query plan structure and frontmatter from disk

bendrucker:
Adds two queries to the session skill that read markdown and YAML from disk through DuckDB community extensions, so plan *structure* becomes queryable instead of regex-parsed. The plan view already exposed plan metadata (`plan_chars`, outcome) from the JSONL index, but answering "which of my plans actually have a Context and Verification section?" still meant reading each file and parsing headings by hand.

`plan-sections` reads plan files with the `markdown` extension and joins each section to the session that produced it via `plan_calls`, carrying outcome and replan sequence onto structural rows. `frontmatter` reads YAML frontmatter with the `yaml` extension.

## Decisions

Plans are read from **disk**, not from the embedded `$.input.plan` string in the JSONL. "Do my plans follow the guidance structure" is inherently a local, current-state question, and the disk path (`read_markdown_sections`) is the clean one. The embedded-string path is blocked anyway: `md_extract_sections(<column>)` can't resolve overloads because the `md` type implicitly casts to/from `VARCHAR`, so both candidates always match. Embedded `$.input.plan` stays the source for `plan_chars` and cross-host rows, so nothing regresses there. A `plan_file` from an imported host (or a deleted plan) simply has no file on disk; the `LEFT JOIN` omits its sections and leaves session-side data untouched.

Extension setup lives in one `resources/extensions.sql`, pulled into a query process via `duckdb -init` (it runs before stdin in the same process, so the piped query sees the extensions, and it works under `-readonly`). Query files stay pure SQL with no `INSTALL`/`LOAD`, and the common-path queries omit `-init` and pay nothing. The index/refresh path stays extension-free.

No view yet. This starts as named queries and promotes to a view only if a second caller appears.

## Deviations From the Plan

Two things the plan assumed didn't hold, and I corrected them while building:

- **`getenv` is CLI-only.** The plan used `getenv('HOME')` for the default glob. That function isn't registered in the `@duckdb/node-api` client the tests (and CI) run on, and it fails at *bind* time even inside an unused `COALESCE` branch, so the query wouldn't even parse in-process. Switched the defaults to a plain `~/...` literal, which DuckDB expands to the home dir and binds in every client.
- **The frontmatter skills default was wrong.** `~/.claude/skills/**/SKILL.md` matches zero files. Real skills are plugin skills under `~/.claude/plugins`, duplicated across cache version-hashes, which makes them a noisy default. Re-defaulted to the memory corpus (clean and frontmatter-native) and documented the skill globs as overrides.

## Testing

The tests in `db.test.ts` load the real `extensions.sql` in-process and cover section parsing, the join to `plan_calls`, and the graceful case where `planFilePath` points at a file absent from disk. I also exercised the `-init` invocation against the live index to check that sections join to their sessions and that the "plans missing a Verification section" query behaves.

The `markdown`/`yaml` extensions are fetched from `community-extensions.duckdb.org` on first `INSTALL ... FROM community` and cached per host after. That host is allowlisted in the sandbox so the first fetch can reach it.

### DIFF
--- a/.claude/rules/settings.md
+++ b/.claude/rules/settings.md
@@ -25,7 +25,7 @@ The sandbox is egress control, not filesystem lockdown. Credentials stay outside
 
 ### Hosts (`WebFetch(domain:...)` Permissions)
 
-- Package registries: `registry.npmjs.org`, `www.npmjs.com`, `pypi.org`, `rubygems.org`, `proxy.golang.org`, `sum.golang.org`.
+- Package registries: `registry.npmjs.org`, `www.npmjs.com`, `pypi.org`, `rubygems.org`, `proxy.golang.org`, `sum.golang.org`, `community-extensions.duckdb.org` (DuckDB community extensions `markdown`/`yaml`, fetched on first `INSTALL ... FROM community`).
 - Docs and source: `docs.anthropic.com`, `code.claude.com`, `modelcontextprotocol.io`, `pkg.go.dev`, `bun.sh`, `bun.com`, `github.com`, `raw.githubusercontent.com`.
 - Credentialed APIs: `api.github.com`, `api.linear.app`, `api.anthropic.com`, `claude.ai`. Trusted only because each secret lives outside the sandbox.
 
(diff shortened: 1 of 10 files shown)


bendrucker:
@claude  Fix the broken test in CI

PR #880, opened 2026-06-24
comments: add slop-comment audit plugin

bendrucker:
Adds a self-contained `comments` plugin that finds AI-slop code comments and trims them. It audits either the comments a change introduced or every comment in a repo, judges each against a two-type model of when a comment earns its place, and writes the deletions to a review branch. The taxonomy and the judge prompt were mined from a real corpus: merged review history and two codebases.

## The Model

A comment earns its place two ways: `what-on-dense` (restating genuinely complex code in words) or `why-on-simple` (explaining a non-obvious reason). The core slop case is `what-on-simple`, simple code with a comment that only restates it. The judge is deliberately permissive. Flagging a good comment is the expensive failure, because a tool that flags good comments gets turned off.

## Scope

One pipeline runs at two scopes, selected by a flag:

- Diff scope (default, `--base <ref>`, `--mr <iid>`): the comments a change introduced, scoped to added and modified lines. In a heavy-GenAI codebase it never touches a teammate's pre-existing comments.
- Repo scope (`--all`): every tracked code file's comments, for sweeping a slop-heavy codebase into one large reviewable deletion.

`--path` narrows either scope. `--sort` and `--limit` rank by intrinsic comment complexity so the longest, densest comments judge first.

Machine-meaningful comments are exempt at collection time: lint and compiler directives (`eslint-disable`, `noqa`, `go:generate`), shebangs, and license headers never reach a judge, since a mistrimmed suppression breaks the build. Generated files and vendored subtrees are skipped the same way.

## Pipeline

Three stages, with a human gate between extraction and judging:

- `preflight` extracts and ranks comments deterministically, builds the judging job on disk, and prints the counts (`N comments / M files / ~K agents / ~T tokens`). The token estimate includes the fixed per-agent overhead (system prompt, tool schemas, the multi-turn cycle), which dominates payload at the default shard size. Nothing fans out until the user confirms.
- A committed Workflow fans out one agent per shard. Agents run pinned to sonnet, the model the eval gate has certified, rather than inheriting the session model. Each agent judges its shard against the rubric and writes verdicts to disk, off the orchestrator context; its structured output is only a total/flagged summary.
- `apply` re-extracts each judged file and matches verdicts to comments by id at their current position, then commits the trims to a fresh `comments/audit-*` branch off HEAD, built with git plumbing so the working tree is never touched. A comment that moved or changed since preflight gets a new id, matches no verdict, and is skipped, so the id match doubles as the drift check. The review surface is a normal `git diff`. `--report` prints the findings instead.

Preflight also writes `features.json` beside the verdicts: deterministic per-comment features (size, comment-to-identifier alignment, why/ticket/divider markers) keyed by the same id. Each run accumulates (features, verdict) pairs, the data for later routing obvious comments away from the judge with measured thresholds. Nothing gates on the features yet.

## Architecture

- `detection/extract.ts` reads comments off Shiki's TextMate tokens (line, block, docstring) and coalesces multi-line blocks via per-line grammar state. Grammars load from `node_modules` on demand, so the plugin vendors no binaries and adding a language is one extension-map entry.
- `detection/collect.ts` unifies the two scopes onto one `CollectedComment`. The diff collector resolves a base (working tree, merge-base, or GitLab MR) and intersects added line ranges with comment ranges, parsing the unified diff with `parse-diff`. The repo collector walks `git ls-files`. Both attach a stable id, the context window, an intrinsic-complexity score, and the deterministic feature vector, and both run the exemption gate (`detection/exempt.ts`).
- `judge/` holds the versioned prompt, the verdict schema, per-verdict validation, and the job builder that shards ranked comments for the Workflow.
- `apply/` is a deterministic edit engine. It deletes full-line comments, strips trailing comments, and trims individual lines of a mixed block, refusing any edit that would break a block delimiter and leaving it for manual handling.

The `comments:audit` skill drives the three stages. It is `disable-model-invocation`, so it carries no recurring idle cost, and a preventive `user/rules/code-comments.md` injects only when a code file is touched.

## Calibration

The judge prompt is a committed, sha256-versioned artifact. `evals/eval.ts` runs it over 22 labeled fixtures through an Anthropic oracle at temperature 0, rendering each batch as escaped XML, and reports a confusion table gated on flagging zero must-pass comments. The oracle is the only place the Anthropic SDK is used. A test enforces that no product code imports it.

The hard case is the ticket-link carve-out. A guard or TODO anchored to a ticket that lifts a present condition (`# extraction_mode is NULL until ENG-2068`) explains a constraint the code cannot show, so it passes. The diary breadcrumb (`arrives with ENG-2065`) is the slop.

Prompt edits change the recorded sha256 and invalidate prior numbers, so the gate must re-clear after any edit. The same applies to the model pin: the fan-out runs the model the gate last cleared, and lowering the tier means re-running `eval.ts --gate` on the cheaper model first.

## Notes

- Personal and local. The audit runs on demand, and the eval needs `ANTHROPIC_API_KEY`.
- The fan-out runs at default temperature, so a malformed trim is possible. The apply engine validates every verdict and skips anything it cannot trim safely, so it under-applies instead of emitting broken syntax.
- Comments can drift between preflight and apply. Apply re-extracts and matches by id, so a changed comment is skipped rather than mistrimmed, and a clean-tree precondition keeps the branch the only mutation.
- Dogfooding against this repo lives in #922, which also surfaced several of the fixes here (heredoc-corrupted verdicts, vendored-tree collection, line-run coalescing).

## Testing

`bun test plugins/comments` covers extraction, the exemption gate, the diff and repo collectors, ranking, the feature vector, the job builder, the edit engine, and the verdict id-match. The live ship gate self-skips without an API key.


bendrucker:
@claude Fix the broken test in CI

PR #918, opened 2026-07-01
jxa: retry once on transient Apple Events error

bendrucker:
The `mac` plugin's `jxa.ts` wrapper runs `osascript` sandboxed, and `sandbox.allowAppleEvents: true` reliably permits the Apple Event. The setting works. But the very first Apple Event after a cold session sometimes fails with a transient macOS XPC hiccup: an `errAEPrivilegeError` (`-10004`), a `-600`, or stderr carrying `Connection Invalid` from `com.apple.hiservices-xpcservice`. These are dispatcher-connection glitches, not real permission denials. A retry clears them.

Previously that transient surfaced as a hard error, which pushed callers to fall back to disabling the sandbox. Now the first call self-heals instead.

`osascript` output is captured (`pipe`) rather than inherited so the wrapper can inspect stderr, then forwarded to the caller unchanged with the real exit code. On a transient first failure the spawn waits a short backoff and retries exactly once. Any non-transient failure, or the result of the single retry, passes through untouched. The loop never runs more than once and never masks a genuine error.

The transient classification lives in an exported pure function, `isTransientAppleEventsError(exitCode, stderr)`. `runWithRetry` takes an injected runner and sleep, so the retry behavior is unit-tested without invoking real `osascript` or a running app: transient-then-success, transient-then-persistent-error, and no-retry-on-non-transient. `validateAppScope` and the cleye CLI are unchanged.

### DIFF
--- a/plugins/mac/scripts/jxa.ts
+++ b/plugins/mac/scripts/jxa.ts
@@ -84,14 +84,15 @@ export interface RunResult {
   stderr: string;
 }
 
-// AppleEvents error codes for transient XPC dispatcher hiccups seen on the first
-// event after a cold session: errAEPrivilegeError (-10004) and errAEEventNotHandled
-// / "application isn't running" (-600). A retry clears them.
-const TRANSIENT_APPLE_EVENTS_CODES = [-10004, -600];
+// AppleEvents error codes that are safe to retry. They surface as transient XPC
+// dispatcher hiccups on the first event after a cold session: errAEPrivilegeError
+// (-10004) and errAEEventNotHandled / "application isn't running" (-600). A retry
+// clears them.
+const RETRIABLE_APPLE_EVENTS_CODES = [-10004, -600];
 
-export function isTransientAppleEventsError(exitCode: number, stderr: string): boolean {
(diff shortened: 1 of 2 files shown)


plugins/mac/scripts/jxa.ts
```diff
@@ -78,6 +78,52 @@ export function validateAppScope(source: string, app: string): ValidationResult
...
+export interface RunResult {
+  code: number;
+  stdout: string;
+  stderr: string;
+}
+
+// AppleEvents error codes for transient XPC dispatcher hiccups seen on the first
+// event after a cold session: errAEPrivilegeError (-10004) and errAEEventNotHandled
+// / "application isn't running" (-600). A retry clears them.
+const TRANSIENT_APPLE_EVENTS_CODES = [-10004, -600];
```
bendrucker: Retriable

bendrucker: Renamed to `isRetriableAppleEventsError` and `RETRIABLE_APPLE_EVENTS_CODES`. Pushed in a420f87's follow-up.

PR #940, opened 2026-07-02
stop: only block on check failures the session caused

bendrucker:
The Stop gate blocks on any prek failure, including repo-wide failures the session did not cause. When `settings-validate` broke against a live upstream schema change, every session that touched a settings file got halted, repeatedly, for days (#860/#869 fixed that instance; a second case hit on 06-24 with `/teammateMode`). This PR classifies each failing check against the session's baseline and stops punishing sessions for failures that predate them.

#### Baseline Capture

Lazy, per check, at the first Stop where that check fails. There is no SessionStart hook and no always-on cost: the pass path adds only a state-file read and a parse of prek's output. When a failing check has no baseline entry yet, the hook resolves the commit HEAD pointed at when the session started, checks it out into a disposable worktree under `tmp/`, re-runs just the failing checks there (scoped to the session files that exist at that commit), and records the verdict.

#### Failure Signature

Check id plus normalized output lines. Normalization strips ANSI, masks durations, and collapses whitespace. A failure is pre-existing only if the check failed at the baseline and every current output line already appeared there, so a pre-existing failure that grows a new error line still blocks. A failure with no output lines matches only a baseline failure that was also silent, since `[].every()` would otherwise cover it vacuously. The verifier rewrites the worktree path to the session cwd before parsing so path-bearing output compares equal.

#### Staleness

Two mechanisms keep a mid-session capture from whitelisting session breakage. The baseline is keyed to git state: HEAD-at-session-start comes from the HEAD reflog (`--date=unix` entry times, which date checkouts by when the ref moved) compared against the transcript's first timestamp, so session commits made before the first Stop are excluded from the baseline checkout. And a check that passes on a later Stop is tombstoned: the session has seen it green, so re-breaking it blocks.

#### Failure Posture

Every verification error (no reflog entry before session start, missing transcript timestamp, worktree add failure, unparseable output, a check without a parsed hook id) yields no verdict, which classifies the failure as new and blocks. The status quo is the fallback in all uncertain paths. Pre-existing failures surface once per session as a non-blocking `systemMessage` and are silent after that.

#### Known Limits

Pre-session uncommitted changes are absent from the baseline checkout, so failures living only in them block as before. Duration masking can equate two failures that differ only in a timing value. A session that fixes and re-breaks a check within one turn produces the same signature as the baseline and passes; the required fixed-then-rebroken case across Stops blocks via the tombstone.

## Review

Ran the code-review skill at medium effort. Fixed from its findings: reflog `%ct` misdating checkouts of old commits (now `%gd --date=unix`), the vacuous empty-output subset, id-less checks poisoning the `prek run` selector, worktree path leakage breaking line comparison, stale worktrees leaked by a killed hook (swept on the next verification), state carried across a session re-rooted into another worktree (state file now keyed by session id and cwd), and an unguarded state write that could throw away a block decision. Added an explicit 300s timeout to the Stop hook entry since the failure path now includes a worktree checkout and the platform default is 60s. Accepted without change: the unconditional `bun install` on the pass path (pre-existing behavior) and blocking when reflog resolution fails for resumed sessions in fresh worktrees (conservative direction).

## Evidence

Local host only. Between 2026-06-18 and 06-24, `settings-validate` failed repo-wide (`user/settings.json: /sandbox: must NOT have additional properties`, schema lagging a live upstream key) and blocked Stop 9-11 times across 4-6 sessions over 6 days. One session was blocked 5 times in 26 minutes. Every other check in those runs passed or skipped, and no block in the window was retried away to success within 300s, so each block was a hard halt on a failure the session could not have caused.

## Testing

Added tests covering the required scenarios at the `evaluateStop` layer and end-to-end through `processStop` with injected prek/verifier/state deps: pre-existing failure notices once without blocking and stays silent after, new failure blocks, pre-existing then fixed then re-broken blocks, mixed new plus pre-existing blocks on the new check while noting the other. Verifier tests run against real scratch git repos: worktree created at the resolved baseline ref, session-created files filtered out, path rewriting, id-less skip, leaked-worktree sweep, and cleanup. Reflog resolution has a regression test for the checkout-of-old-commit case. Also exercised the real hook against this worktree: a passing run stays silent and a fabricated new failure blocks with verification running in a real baseline worktree, leaving no leftover worktrees.

Discovery: 83ab08f815d3

https://claude.ai/code/session_01LmBorNSEbodhWp4ndghgjy

### DIFF
--- a/.claude/hooks/stop/baseline.test.ts
+++ b/.claude/hooks/stop/baseline.test.ts
@@ -1,0 +1,457 @@
+import { afterAll, beforeAll, describe, expect, it } from "bun:test";
+import { mkdtemp, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import {
+  type BaselineVerifier,
+  type CheckResult,
+  createVerifier,
+  evaluateStop,
+  normalizeLine,
+  parsePrekOutput,
+  readState,
+  resolveHeadAt,
+  type StopState,
+  transcriptStartTime,
+  writeState,
+} from "./baseline";
+
+let tempDir: string;
+
+beforeAll(async () => {
+  tempDir = await mkdtemp(join(tmpdir(), "stop-baseline-test-"));
+});
+
+afterAll(async () => {
+  await rm(tempDir, { recursive: true, force: true });
+});
+
+const SAMPLE_OUTPUT = `Plugin tests.............................................................Passed
+Plugin validation....................................(no files to check)Skipped
+Settings validation......................................................Failed
+- hook id: settings-validate
+- exit code: 1
+
+  user/settings.json: /sandbox: must NOT have additional properties
+Marketplace completeness.................................................Failed
+- hook id: marketplace-check
+- exit code: 1
+
+  Plugins missing from marketplace.json:
+    bogus
+`;
+
+describe("normalizeLine", () => {
+  it("strips ANSI escapes and collapses whitespace", () => {
+    const esc = String.fromCharCode(27);
+    expect(normalizeLine(`  ${esc}[31mfail${esc}[0m   here `)).toBe("fail here");
+  });
+
+  it("masks durations", () => {
+    expect(normalizeLine("Ran 3 tests. [12.34ms]")).toBe("Ran 3 tests. [<duration>]");
+    expect(normalizeLine("timed out after 5 s")).toBe("timed out after <duration>");
+  });
+});
+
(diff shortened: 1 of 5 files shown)


bendrucker:
Closing as over-complicated per review. The finding stands (a pre-existing repo-wide check failure re-blocked 4+ sessions over 6 days) and is recorded in the W27 discovery digest; if it recurs, revisit with a simpler mechanism than baseline worktree re-verification.

PR #1034, opened 2026-07-10
session: move index to plugin data dir with incremental per-file imports

bendrucker:
The session index never actually lived in plugin data. `CLAUDE_PLUGIN_DATA` is a skill-text template and a hook/MCP env var, but every real invocation of `refresh.ts` happens in a Bash tool shell where it is empty, so the TMPDIR fallback was the de facto production location: a 33GB copy in `/tmp/claude-session` (plus a divergent TMPDIR twin), rebuilt from scratch after every macOS purge, with the per-session freshness marker never firing because `CLAUDE_SESSION_ID` is empty in Bash too. This PR fixes the harness around the queries: placement, growth, freshness, change detection, and the documented lock model.

## Decisions

`getDataDir` derives the data dir from the script's own installed location (`plugins/cache/<marketplace>/<plugin>/...` maps to `plugins/data/<plugin>-<marketplace>/`) rather than hardcoding the first-party id, keeping `CLAUDE_PLUGIN_DATA` as the override. A bare repo checkout errors instead of falling back, so development can never touch the live index. You chose this over hardcoding `claude-code-bendrucker` and over threading the template through every documented command.

Change detection moves from a per-host mtime watermark to an `indexed_files(host, path, mtime, size)` catalog. The watermark had two holes: a write landing between scan and stamp was skipped forever, and `rsync -a` delivering a new file with a preserved old mtime was never indexed. The catalog also drops rows for deleted files, with one guard: a missing host root means an unmounted or misconfigured path, so the scan skips that host rather than dropping all its rows, and a typo'd `CLAUDE_PROJECTS_DIR` cannot wipe a host.

Imports now run per changed file (`DELETE` + `INSERT` in a transaction) instead of rewriting `raw` and `content_items` wholesale, and every writing run ends with `CHECKPOINT` plus a real `DuckDBInstance` close. DuckDB only reuses freed blocks after a checkpoint and never shrinks a file, which is how a 979MB corpus produced a 33GB index. A compaction guard in `refresh.ts` (file > 4x corpus, 64 MiB floor) rebuilds via `COPY FROM DATABASE` with an atomic rename for recovery from pathological bloat. `source_line` stays SQL-computed but per-file (`ROW_NUMBER() OVER ()` on a single-file scan, no cross-file partition); the formal-undefinedness caveat is documented in Source Lookup.

Freshness is a `last-refresh` stamp with `--max-age` (default 300s): the fast path prints the path and exits without opening the database, so it takes no lock and coexists with running readers. Session markers and the `CLAUDE_SESSION_ID` dependency are gone. `views.sql` is fingerprinted in `index_meta`, so editing a view takes effect on the next refresh even with no changed files.

SKILL.md now documents the real DuckDB lock model instead of claiming read-only opens take no lock: readers share a lock and coexist, a writer needs exclusivity, and collisions fail with `Could not set lock` to be retried. The stable `${CLAUDE_PLUGIN_DATA}` path is stated once, removing the `$DB` command-substitution threading from the fan-out recipes in `references/discovery.md` and `improve-claude-code`. Cross-machine import/re-sync/forget mechanics move to `references/cross-machine.md`, cutting SKILL.md's always-loaded cold path. The stray TMPDIR indexes are swept by `refresh.ts` on production runs only (an env override marks a test/dev run, which must not delete machine-wide state).

## Testing

Unit and integration suites extended: rsync-gap (old-mtime new file gets indexed), deleted-file row drops, missing-root guard, view-fingerprint rebuild and skip, `source_line` 1..N per file, compaction preserving tables/views/macros, TTL fast path vs `--refresh`, and bounded file growth across five reimports.

Verified end-to-end: a copy of the skill placed in a fake `plugins/cache/bendrucker/claude-code/<hash>/` layout derived the data dir and built the index with no env vars; a bare-checkout run errored as designed. A full production build against the live corpus (991MB local + 1.6GB imported host) produced a 5.8GB index at `~/.claude/plugins/data/claude-code-bendrucker/`, warm-path reruns complete in 0.04s while a concurrent `duckdb -readonly` holds the file, and cross-host queries return 861k rows. This session also removed the 33GB `/tmp/claude-session` stray.


bendrucker:
@greptile-apps review

PR #1037, opened 2026-07-10
rate-limits: burn-rate usage prediction plugin

bendrucker:
Adds a `rate-limits` marketplace plugin whose alert is linear burn-rate prediction. Claude Code exposes usage only through the statusline `rate_limits` payload, and only two fields per window: an integer `used_percentage` and an epoch-seconds `resets_at`. The prior band-crossing approach (the `session-limit` hook) could only warn once you were already near the ceiling at 90/95/100%. It could not tell you at 40% that your current pace will exhaust the block before it resets. This plugin fits recent `(time, used_percentage)` samples within the current 5-hour block, projects when the line reaches 100%, and warns when that lands before `resets_at`.

## Prediction

`evaluate` fits a least-squares slope over the in-block samples (those sharing the current `resets_at`), projects `t* = now + (100 - pct)/slope`, and fires when `slope > 0` and `t*` lands more than `marginMs` before reset. Fitting is guarded against integer-percentage noise: a minimum sample count, a minimum time spread, and a positive slope are all required before a one-percent tick can read as a rate. The 100%-exhausted terminal guidance (stop, schedule a wake-up if the reset is within the auto-schedule horizon) is kept as a prediction-independent backstop. The 90/95 "favor efficiency" bands are dropped.

## History and the Recorder

Prediction needs history the single-snapshot `rl.json` cannot hold, so the plugin ships a recorder: a statusline wrapper that reads stdin once, mirrors `rate_limits` to `rl.json` (single writer, so Vibe Island is unaffected), appends an edge-triggered NDJSON history line, then execs the inner statusline. History is edge-triggered (a line only when the observation changes), pruned to 7 days on each write via best-effort full rewrite, and collapses concurrent-session duplicates on read. Recording is best-effort so rendering never blocks.

## Compile and Dispatch

`hooks.json` uses a shell dispatcher that prefers a compiled binary at `bin/alert` and falls back to `bun` when it is absent. The `install` command runs `bun build --compile` to produce it. The binary is gitignored and ephemeral by design: a plugin upgrade lands in a fresh root with no binary, and the dispatcher falls back to `bun` until the skill recompiles, so no version-stamp or orphan-cleanup machinery is needed.

## Surfaces

- **Skill** (`rate-limits`): install/prepare, doctor, status, and a model-callable pre-flight check before large work.
- **`doctor`**: verifies `rl.json` (path, exists, parses, numeric percentage, freshness), history (exists, fresh, enough in-block samples), the compiled hook, and the statusline recorder, each with remediation.
- **`status`**: current percentage and reset per window plus the projection, and stamps the marker for parity with the hook so a query does not leave a redundant alert queued.

The marker dedups per session, keyed to the block via `resets_at`. It re-arms on block rollover, and a materially earlier projection re-alerts within a block while an unchanged one stays quiet.

## Terminology

The plugin stays named `rate-limits` and reads the literal `rate_limits` wire field, but all user-facing prose says "usage limits", Anthropic's product term for the 5-hour and weekly limits.

## Verification

- Unit coverage spans prediction (a steepening series signals while a shallow one that clears reset stays silent, the noise guards suppress fits, cross-surface dedup, materially-worse re-alert, block-rollover re-arm, exhausted backstop), the edge-triggered history bound (unchanged percentage adds no line, the 7-day prune drops stale lines, read-side dedup), and doctor (fresh-and-populated passes; missing, stale, or undersized history fails with remediation).
- Exercised end-to-end: the recorder (mirror, edge-trigger, inner exec), `status` (projection and marker stamp), `doctor` (healthy pipeline), and `install` (compiled `bin/alert`), confirming the hook emits the prediction identically via bun, the compiled binary, and the dispatcher.
- Marketplace listing, skill-lint, biome, plugin-import, and hook-quoting checks are covered.

This ports the terminal backstop and marker from the WIP scaffold and does not migrate the personal `user/` setup onto the plugin. That is a follow-up once the plugin proves out.


bendrucker:
Don't have time to work on this right now, threshold is ok.

PR #1059, opened 2026-07-15
site: add a plugin marketplace browser

bendrucker:
Adds a static site under `site/` that browses the marketplace's 33 plugins, built for `bendrucker.github.io/claude/`. Nothing browsed them before and the root README has no install instructions, so discovery meant reading directories on GitHub.

Preview locally with `bun run site:dev`, which serves at `http://localhost:4321/claude/` (the `/claude` base applies in dev too).

## Scope

One page per plugin, 35 pages total. The README is the only markdown the site renders. Skills, agents and hook events are frontmatter and metadata shown as elements on that page, each linking to its source. Viewing source on GitHub is always an option, so the site summarizes and leaves code browsing to GitHub, which already does it well.

Hooks summarize to one element per event. A plugin can back an event with several matchers and scripts (`tmux` registers four events, two of them matcher-less), and the summary collapses that rather than reproducing the file.

Component chips come from one `COMPONENT_TYPES` table, so a new upstream type is one row. An empty array renders no chip, which is why `go` and `newline` honestly show hooks alone, and why the two remote plugins (`agents-md`, `honeycomb`) show a source repo and nothing else while staying installable.

## Plugin age follows renames

`site/src/data/history.ts` derives each plugin's age from one whole-history `git log` pass. The lineage rule: a rename A to B counts only if A no longer exists at HEAD. That follows `implement-issue` and `refine-issue` into `issue`, and `style` into `writing`, while rejecting the `claude-code` to `writing` file migration between two plugins that both still exist. Without the rule `writing` reads 75 days too young and `issue` 48.

Two things that look like details and are not. Commits sort by log position rather than by comparing ISO strings, because this repo mixes `-08:00`, `-07:00` and `Z` offsets. Dates stay author-local, because converting `issue`'s first commit to UTC moves it a day.

The module throws in a shallow clone. `actions/checkout` defaults to `fetch-depth: 1`, which yields empty dates while the build still succeeds, so a future edit dropping `fetch-depth: 0` from the site jobs stops the build rather than shipping bogus history.

## Privacy

`loadPlugins()` defaults to reading the real `user/settings.json` for `enabled`, and `Plugin.dir` is an absolute `/Users/ben` path. Both would otherwise reach the published `plugins.json`. The catalog type omits them structurally and entries are built field by field, so re-adding either fails to compile. `loadPlugins()` is called with a settings path that cannot exist. A test asserts the serialized catalog matches neither `/Users/` nor `"enabled"`.

## Categories

`category` and `keywords` are already in the upstream SchemaStore marketplace schema, so they needed no `schemas/overlays/` change and `packages/validate/marketplace.ts` covers them for free. `packages/marketplace` gains only a conditional passthrough for the two fields.

## Two things the build would have gotten wrong

`bun --bun run --cwd site <script>` does not force Bun's runtime. `--bun` applies when `run` gets a file path; given a package.json script name it shells out and silently runs Node, which breaks every `Bun.file` read the build depends on. The root `site:*` scripts point at `node_modules/.bin/astro` instead.

`loadCatalog()` resolves its root through a Vite `define` rather than `import.meta.dirname`. Astro's build flattens SSR modules into one chunks directory, so a walk-up by fixed depth lands outside the repo. Unbundled `bun test` runs never see this.

## Verification

Beyond what CI posts:

- The built `plugins.json` carries `issue` at 2026-01-09 and `writing` at 2026-01-13, against the naive 2026-02-26 and 2026-03-29. `writing` does not inherit `claude-code`'s history.
- A `--depth 1` clone fails the history load rather than building with empty dates.
- No relative `.md` link survives into the built HTML, and no internal href escapes `/claude`.
- Checked the index, a plugin page, the modal and the dependency chain in a browser. Opening `issue`, clicking through to `pull-request` and pressing Back walks back through the chain to the index.

`plugins.json` is a prerendered endpoint serializing the same object the pages consume, versioned with `schemaVersion: 1`, so another site can consume it without a cross-repo trigger and it cannot drift from what renders.

The marketplace has 7 plugin agents rather than the 6 I expected going in: `type-ignore/fixer`, `github/{logs,rulesets-manager}`, `gitlab/logs` and `writing/{content,artifacts,style}`.

## Before merge

Pages is not enabled on the repo. Enabling it is a one-time step needing repo-admin:

```
gh api --method POST repos/bendrucker/claude/pages -f build_type=workflow
```

`pages.yml` deploys on push to `main` with no path filters, since the content derives from `plugins/**`, `marketplace.json`, `site/**` and the commit graph, which has no filterable surface.


bendrucker:
Closing this as stale. Main has moved 41 commits past this branch and the marketplace site would need rebuilding against the current plugin layout rather than merging. Reopen or start fresh if the plugin browser comes back on the list.

PR #1105, opened 2026-07-24
skills: fix forked skills whose tools are unavailable in a fork

bendrucker:
`interview:plan` and `interview:clarify` mandate `AskUserQuestion` in every step while running `context: fork`. A forked skill is a regular subagent, and `AskUserQuestion` is stripped from every subagent. Neither skill can do the one thing it exists to do. A run on 2026-07-15 returned "the tool is not available in this agent context (confirmed via ToolSearch)" and degraded to a prose question list, which both skills explicitly forbid. Removing `context: fork` runs them inline, where the tool exists. `agent: Plan` is meaningless without the fork and goes with it.

`background: false` cannot fix this. The `AskUserQuestion` removal is the first subagent filter and applies foreground or background.

`writing:rewrite` keeps its fork and gains `background: false`. Forks background by default, and its rewritten output lands in a later turn while the user is still waiting on it before sending a message. The justification here is turn timing. Both of its declared tools (`Bash`, `Read`) are in the background subagent set, leaving nothing degraded.

Documents `background` in the skill-authoring reference, which never mentioned it, and corrects `patterns.md`, which described forks as if they still block the turn.

Accepted tradeoff: the interview skills now run inline and inherit the session's permission mode. The prose write-guards at `plan/SKILL.md:38` and `clarify/SKILL.md:27` become the only guard against writes during an interview. Their exploration also lands in the main context window instead of a throwaway fork.

`background` is the repo's first use of that key. It needs no schema entry: #1109 deleted `schemas/skill.schema.json`, and rebasing onto that leaves no frontmatter schema to keep in sync.

https://claude.ai/code/session_01JwSBM8AdjwpMgZPhzJ9dKx

### DIFF
--- a/plugins/claude-code/skills/skill/SKILL.md
+++ b/plugins/claude-code/skills/skill/SKILL.md
@@ -52,6 +52,7 @@ model: sonnet                             # Optional: override model
 effort: low                               # Optional: reasoning effort
 context: fork                             # Optional: run in isolated subagent
 agent: Explore                            # Optional: agent type for fork
+background: false                         # Optional: block the turn on a fork
 user-invocable: false                     # Optional: hide from slash menu
 hooks:                                    # Optional: skill-scoped hooks
   PreToolUse:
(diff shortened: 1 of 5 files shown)


bendrucker:
Moot after rebasing onto current `main`. #1109 deleted `schemas/skill.schema.json` outright, so there is no `additionalProperties: false` left for `background` or `argument-hint` to violate, and `grep` finds no remaining reference to the file. That PR removed the schema for exactly the reason this finding describes: it had no consumer and had drifted from the frontmatter it claimed to specify.

Body updated to drop the deferral note.

PR #1155, opened 2026-08-06
hooks: fix matchers that could never fire

bendrucker:
The pull-request plugin's body-validation hook has never fired. Its PreToolUse matcher used permission-rule syntax (`Bash(gh pr create:*)|...`), but a matcher is a regex over tool names, so the pattern could never match `Bash` and the dispatcher never ran `validate.ts`. The session index shows zero live firings, and the body of #1142 went through despite containing a test count the hook denies. Dispatch now uses `matcher: "Bash"` with one permission rule per command form in the `if` field, which evaluates per subcommand with env prefixes stripped, so compound and env-prefixed invocations match too.

The same defect class was live in eight more hooks, all found by a new CI lint (`scripts/check-hook-matchers.ts`) that rejects three shapes the dispatcher parses as never-matching:

- permission-rule syntax in a matcher: git, linear, shortcuts, tmux
- pipe-joined `if` rules, where the docs require exactly one: go, type-ignore
- URL-form WebFetch rules, where only `domain:` matches: github, gitlab

All eight are rewritten to the working shape with behavior otherwise unchanged. Reviving them turns on gates that were silently off, most visibly git's block on committing to the default branch.

Validation strictness now splits by fixability. Findings with an exact mechanical fix deny: test counts, heading case (the reason carries the corrected headings), and backticked refs or SHAs. Judgment calls stay warnings: scaffold, file tours, CI roll-calls, run-on prose. A deny folds the warnings into its reason so one edit clears everything, instead of a second blocked attempt discovering them.

Unit tests could not have caught the dead matcher: they exercise the script, not the dispatcher. The new `e2e-hook-fire.ts` runs headless `claude -p` with `--plugin-dir` against a throwaway repo, with a stub `gh` on PATH recording invocations. A fixture with a test count must be blocked before the stub runs, with the deny reason visible in the stream output, and a clean fixture must reach the stub. Both held in a real local run.

The `hook-e2e` workflow repeats that proof in CI: it discovers every `plugins/*/scripts/e2e-*.ts` and runs them all on PRs touching any plugin's hooks or scripts, so a new plugin's test needs no workflow edit. It authenticates with the existing `CLAUDE_CODE_OAUTH_TOKEN` secret, since each run spends API tokens and the repo has no `ANTHROPIC_API_KEY` secret. I did not extend dispatch proof to the other eight plugins. The lint guards the class statically.

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -17,6 +17,7 @@ After making changes to plugin scripts, run them directly to verify they work en
 - **Prefix dotdir paths with `./` in `bun test`**: A positional arg is a *filter* (substring match against discovered test paths), not a path. Discovery skips dotdirs, so `bun test .claude/hooks` matches nothing and silently runs 0 tests. Writing `bun test ./.claude/hooks` makes bun read it as a path and run the tests under it. Always use a `./` prefix for hidden-dir paths in CI and local test commands.
 - **No `.js` imports in TypeScript**: Import from `./module` not `./module.js`. The bundler/runtime handles resolution.
 - **Prefer skills over agents**: Skills are invocable via the `Skill` tool. Agents require the `Agent` tool. If something should be directly invocable, make it a skill.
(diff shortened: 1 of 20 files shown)


bendrucker:
The hook-e2e failure is the repo's CLAUDE_CODE_OAUTH_TOKEN secret, not the code: 401 OAuth access token has expired. This job is the first in CI that makes a real API call (context and plugin validate never leave the runner), so the expiry went unnoticed until now. The same script passes locally against live auth. To fix: `claude setup-token`, then `gh secret set CLAUDE_CODE_OAUTH_TOKEN`, then re-run the job.

PR #1160, opened 2026-08-06
herdr: port CLI semantics from upstream's agent skill

bendrucker:
Ports the behavior herdr's own upstream agent skill documents into `user/skills/herdr/SKILL.md`, rather than adopting the upstream file (`skills/herdr/SKILL.md`, installed via `npx skills add herdrdev/herdr`) directly.

I didn't adopt it outright. Upstream's description restricts activation to when the user says "herdr" explicitly, which conflicts with this skill's situational design: it needs to activate on a background terminal, a handoff, or parallel work without that word appearing. Installing it globally would also write through `~/.claude/skills/herdr` into this repo's own tracked copy.

This repo's command surface is already generated from the installed binary's `--help` at load time. It already carries every flag and enum value upstream states in prose. What `--help` can't state is behavior, and that's what I ported: `--current` targeting, the `idle`/`done` split on whether a tab has been seen, `agent start`'s pane precondition, the alternate-screen scrollback ceiling, and `wait-output` searching the current snapshot before it waits.

I checked every claim against the installed 0.8.0 binary rather than copying it from upstream's prose. Two of upstream's own descriptions were wrong. `wait-output` was described as only blocking on new text appearing. The plugin section pointed at a platform filter flag that doesn't exist; the skill now uses the `jq` filter that does the same job instead.

`pane`, `tab`, `workspace`, and `split` are now herdr's own terms in the skill's description, so a request naming one activates the skill without saying "herdr". `tmux:tmux` also claims "pane". Nothing at description-match time can see `HERDR_ENV`, so a line in the body resolves the ambiguity once the skill has loaded.

Deferred: the `reviewr` example's comment header hardcodes a path into this same file, which drifts on every edit even though the surrounding text calls the snippet the reliable anchor and the header only a hint.

### DIFF
--- a/user/skills/herdr/SKILL.md
+++ b/user/skills/herdr/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: herdr
 description: >-
-  Drive the herdr terminal workspace manager: inspect workspaces, tabs, and panes, hand work to sibling coding agents and read their results, split panes for collaborative file viewing or long-running processes, and correlate panes to Claude sessions. Use when coordinating with another agent, opening a file alongside the user, starting a dev server or log tail the user should watch, capturing another pane's output, or asking what else is running.
+  Drive the herdr terminal workspace manager: inspect workspaces, tabs, and panes, hand work to sibling coding agents and read their results, split panes for collaborative file viewing or long-running processes, and correlate panes to Claude sessions. Use when coordinating with another agent, opening a file alongside the user, starting a dev server or log tail the user should watch, capturing another pane's output, or asking what else is running. Pane, tab, workspace, and split are herdr's terms, so a request naming one is a herdr request even when it never says herdr.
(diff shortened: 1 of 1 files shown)


bendrucker:
Added a commit folding in the [agent-automation doc](https://herdr.dev/docs/agent-automation/) semantics. The main correction: the alternate-screen paragraph claimed scrolled-away rows are unrecoverable, but that only holds for `pane read`. `agent read` on a resting recognized agent pages history out through the agent's own mouse-scroll interface, verified live (a 300-line read returned 92 rows on an idle agent). Deep reads during working/blocked/unknown come back truncated or as `agent_not_idle`. Also picked up: `--until` repeats for multiple states and requires `--wait` on `prompt`, the pane-vs-agent primitive distinction, and `pane send-text`.

PR #1165, opened 2026-08-08
tmux: prune dead hook registrations

bendrucker:
Removes three tmux hook registrations that have zero fires in the plugin's entire lifetime, per session-corpus evidence, while sibling paths in the same plugin fire fine (`notification.ts --background-only`: 15 fires):

- The `Notification` hook's `permission_prompt` matcher
- The `PreToolUse` `target.ts` registration (`Bash(tmux:*)`)
- The `SessionStart` `@resume-command` hook

The user is mid-migration from tmux to herdr, so this trims only the never-fired registrations. The plugin and its working paths (`context.ts`, and `notification.ts`'s `--clear` and `--background-only` paths) stay as-is. Removing the `target.ts` registration leaves nothing else referencing `target.ts` or `target.test.ts`, so this PR deletes both.

**Merge-order note**: open PR #1161 modifies `plugins/tmux/hooks/target.ts`, hardening `parseTmuxCommand` against multi-line commands. This PR deletes that file entirely, so the two will conflict. Since #1161's hardening addresses a registration this PR removes for being dead, land this PR first and drop the now-moot `target.ts` changes from #1161, or land #1161 first and rebase this deletion on top.

Discovery: d81fefb5acba


bendrucker:
Rebase note: the `target.ts` deletion's evidence is stale as of #1155. Its zero fires were measured while the entry's matcher was the broken `Bash(tmux:*)` form that #1155 just repaired, so the corpus measured a bug, not a useless hook. The other two prunes (`@resume-command`, the `Notification` `permission_prompt` entry) had working matchers and their zero-fire evidence stands. Worth deciding whether `target.ts` gets a second look before this merges.

PR #1177, opened 2026-08-08
hooks: delete if: conditions that never matched

bendrucker:
Deletes the `if:` conditions on the biome, type-ignore, and go hooks. Written as file-path globs (`Edit(**/*.ts)|Write(**/*.ts)`), they have never matched a single edit across 2,173 `.ts` and 1,249 `.go` edit operations since configuration. Bash-prefix `if:` conditions on the same hook infrastructure fire fine. All three scripts already self-filter by file extension, so the broken glob condition was dead weight sitting in front of a filter that already did the job.

The coverage `PostToolUse` hook shared the same broken `if:`. Unlike the other three, it has zero invocations to show for it even accounting for the bug. The `coverage` skill exists only to react to its hint, so a hook that never fired means the skill never got used either.

- Removes the hook entry from `.claude/settings.json` and the `.claude/skills/coverage/` skill directory.
- Removes the now-orphaned `.claude/hooks/coverage/` script, since nothing else referenced it once the settings.json entry was gone.
- Removes the two `.gitignore` re-include lines that existed only to un-ignore those two paths against a global `coverage` ignore rule.
- Leaves `scripts/coverage/` (the underlying `bun run coverage` engine, still used standalone in CI) untouched.
- Swaps a test fixture in `check-skill-invocability.test.ts` that used the deleted skill's path as an example.

Hook config loads once at session start, so the running session that made these edits couldn't observe them taking effect. Verified with a fresh `claude -p` session in this worktree instead. Editing a `.ts` file with a duplicate object key and a `@ts-ignore` surfaced both the biome and type-ignore `additionalContext` blocks. Editing a `.go` file carrying Go's generated-code marker comment was blocked by the go `PreToolUse` hook, which previously let it through silently.

Discovery: 06f0afe8e38e
Discovery: e43045037d91


bendrucker:
Rebase note: main's #1155 repaired the same `if:` globs this PR deletes (splitting the pipe-joined conditions into per-glob entries that dispatch correctly). The original evidence that these conditions never matched was measured against the broken form. Both endpoints behave correctly since the scripts self-filter by extension, but keeping #1155's fixed entries would skip the process spawn on non-matching files, while this PR's deletion spawns the hook on every Edit/Write and filters inside. If per-glob dispatch is worth keeping for the spawn savings, the go and type-ignore hunks should be dropped from this PR. The `.claude/settings.json` biome `if:` removal and the coverage deletions are unaffected.

PR #1178, opened 2026-08-08
user: nightly launchd worktree prune

bendrucker:
Adds a nightly launchd job that runs `wt step prune` across every repo under `~/src`, cleaning up worktrees and branches already merged into their default branch.

The sandbox's worktree deny-list scales with worktree count, and posix_spawn has a fixed argument-size limit. Overflowing it (E2BIG) has already killed every Bash and Skill spawn for the rest of a session once (155 errors across 13 sessions). The current inventory (~25 in-repo `.worktrees` plus 19 sibling worktrunk dirs) sits close to the recorded ~30 ceiling, with nothing pruning it back down. `tool_errors ILIKE '%E2BIG%'` in the session index is the signal to watch for recurrence.

`wt step prune` (not `wt prune`, which doesn't exist as a top-level command) already guards two things. It refuses to touch a dirty worktree without `--force`, which this job never passes. It skips worktrees younger than `--min-age`, left at the 1-day default.

It doesn't guard a third case: a worktree that's clean and merged but still has a live process or interactive shell `cd`'d into it. `wt remove --reap` can kill processes during an explicit `wt remove`, but that's opt-in, not automatic on `wt step prune`. It only handles processes, not an idle session sitting in a directory about to disappear.

I'm flagging this as a Worktrunk feature request (a refusal, not just `--reap`, when a live process's cwd is under a prune candidate) instead of working around it. The realistic exposure is narrow: uncommitted work is already protected. 03:15 is chosen to further minimize the odds of catching an active session.

Ran a Greptile local review before pushing. It flagged one P2 readability nit (a semicolon-dense paragraph in the README), fixed in the second commit.

Discovery: b338ed977bc7


bendrucker:
Hold before merging: this duplicates a nightly prune that dotfiles already deploys. `~/Library/LaunchAgents/com.user.worktree-prune.plist` (installed Jul 23) runs `~/.dotfiles/bin/worktree-prune` at 04:00 daily, which wraps `wt-all prune --before`, reports failures into Things, and audits for prune drift. It ran green this morning (107 repos scanned). The discovery finding that motivated this PR predated knowledge of that job.

Decision needed: either close this in favor of the dotfiles job, or adopt this as the replacement and remove the dotfiles one. The E2BIG recurrence-signal documentation in the README is worth keeping either way.

PR #1221, opened 2026-08-17
things: serve the MCP tools over stdio

bendrucker:
Recovers the Things MCP server lost with the unmerged `fix-1049` branch and converts it from Streamable HTTP to stdio, so [tailgate](https://github.com/bendrucker/tailgate) can spawn it as a child process. Things is tailgate's first production upstream, which is what makes it reachable from a phone or a laptop. Replaces the `mcp-gate` and `things-mcp` launchd pair.

`server.ts`, `gate.ts`, `auth.ts`, and `install.ts` are not recovered. tailgate owns OAuth, token introspection, audience validation, and per-identity policy, so this server has no auth code, no port, and no socket.

## Stdout Discipline

This is why the change reaches outside `src/mcp/`. On stdio, stdout is the JSON-RPC channel, so a print from any module in the import closure corrupts the framing. The server reuses the plugin's CLI scripts as modules, and two of their writes sat on the tool path:

- `reorder()` logged its result and called `process.exit(1)`. It returns a `ReorderResult` and throws instead, leaving both to the CLI.
- `openUrl` ran Bun's `$`, which inherits stdout. It runs `.quiet()` now and forwards what it captured to stderr, where tailgate logs it.

Nothing outside `plugins/things` imports these scripts, since `check-plugin-imports.ts` forbids it. The shared surface is this plugin's own `url`, `triage`, and `jxa` skills invoking them as CLIs. Their CLI behavior is unchanged.

`stdout.test.ts` guards the invariant: it walks the import closure and snapshots every reachable `console` call, `process.stdout` write, and unquieted `$`. The handshake test drives only `initialize` and `tools/list`, so it never enters a tool handler and cannot see a print added inside one.

## Review Fixes

Four defects on the write path:

- `.quiet()` captured `open`'s diagnostics and dropped them, since a `ShellError`'s message is only its exit code. A failed handoff reported `Failed with exit code 1` and nothing else.
- `buildAttribution` defaulted its directory to `process.cwd()`, which under tailgate is the server's spawn directory. A `capture_inbox` carrying `session_id` alone produced a resume snippet pointing somewhere unrelated. The caller names the directory now, and one that cannot gets a snippet with no `cd`.
- A batch update failing partway discarded which batches had already applied, so a retry would update those todos twice.
- Blank titles reached Things as nameless todos.

Both runner resolvers duplicated the same two-layout probe. #1220 fixed the installed-layout scan in one (it gated on `Bun.file(<dir>).exists()`, false for a directory, so no installed plugin ever resolved a runner) and `jxa.ts` carried the same defect. `findSiblingScript` is now the single implementation.

## Verification

A live handshake against real Things data returns three JSON-RPC lines: `initialize`, `tools/list` with all nine tools, and a `list_todos` read of today. Empty stderr, no stray stdout. Tests cover the resolver across both marketplace layouts, the stdout closure, the reorder guards, the attribution footer, and the tool registry.

Write tools stay unexercised, since they mutate real task data. Unit tests cover their validation and parameter building.

The plugin `tsconfig.json` reports a pre-existing `moduleResolution=node10` deprecation. Switching it to `bundler` surfaces five `checkJs` errors in the JXA scripts, so that is its own change.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -207,8 +207,10 @@
       "version": "1.0.0",
       "dependencies": {
         "@anthropic-ai/claude-agent-sdk": "^0.3.0",
+        "@modelcontextprotocol/sdk": "^1.30.0",
         "cleye": "^2.0.0",
         "table": "^6.9.0",
+        "zod": "^4.4.3",
       },
       "devDependencies": {
         "@jxa/global-type": "^1.4.0",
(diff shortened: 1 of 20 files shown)


bendrucker:
@greptileai review

PR #1224, opened 2026-08-18
github: govern the copilot review by budget

bendrucker:
The `github:copilot` skill was `disable-model-invocation`, so the only way in was a slash command I don't type, and the session index records zero invocations ever. Every credit I actually spent came from ad-hoc CLI runs with no budget guard and no model default. The skill becomes model-invocable, and a credit meter replaces the invocation barrier as the governor.

#### Budget Guard

Preflight reads `gh api /copilot_internal/user` and derives a tier from `pace`, credits remaining over days to reset: constrained under 25 a day, abundant over 60. The tier sets how strict the bar is, never a spend target: a target would invite manufacturing reviews to hit it.

The script refuses `--agentic` outside the abundant tier, degrades to `gpt-5.6-luna` below 150 remaining, and refuses any run whose session cap plus an overshoot reserve would pass the 1500-credit entitlement. Every spawn carries `--max-ai-credits`. A meter it cannot read stops the run, which keeps this inert on a machine with no personal Copilot entitlement.

Copilot's session cap is soft: a session stops only once it observes usage at the cap. The reserve therefore scales with the session count, because three angles are three sessions that can each overrun by a response. That puts the 3-angle wall at 1335 used, stranding up to ~165 credits in a month that reaches it.

#### Shapes

`--agentic` hands Copilot a disposable worktree at HEAD and tools, so it reads callers, tests, and history instead of inlined text. One capped session at 60 credits, `--allow-all-tools` minus a deny list: `write`, `url`, `shell(git push)`, `shell(gh:*)`, `shell(curl:*)`, `shell(wget:*)`. It needs a clean tree, or the checkout would not match the diff it was handed.

#### Decisions

A probe settled the caching question first: two terra spawns sharing an identical 20.5k-token prefix each wrote 34.6k tokens with `cache_read_tokens` at zero. No prefix cache survives across processes, which killed a planned prompt reorder and made serializing the angles nearly free. `HOME` now persists at `~/.cache/claude/copilot-home`, so those spawns write one auditable session ledger.

Both prompt shapes now read one set of defect-class definitions. The single-angle default had carried a hand-copied duplicate that drifted weaker than `--angles 3`.

#### Ship Gate

`/ship` gains a Cross-Model Gate: code changed on a repo I own, clearing the same list the Bot Review Gate uses. `--skip copilot` turns it off for a run. The two gates share criteria and not budgets, since Greptile's free tier never resets while Copilot's refills monthly.

Removal trigger: a billing cycle with near-zero `skill_calls` for `github:copilot`.

#### Known Gaps

The guard holds no lock, so simultaneous reviews reserve against one reading. A separate $0 stop-usage budget on All AI Credit SKUs is the backstop.

Agentic containment holds against misbehavior rather than an attacker. `--deny-tool write` leaves shell redirection open, so the worktree is the write barrier, and it shares the parent's `.git`. Prompt injection is handled by a usage rule: agentic reviews changes I wrote, and untrusted diffs get the one-shot shape.


bendrucker:
@greptileai review

bendrucker:
@greptileai review

PR #1234, opened 2026-08-19
session: add a sandbox-bypass justification query

bendrucker:
Adds `sandbox-bypass-justification`, a named session query reporting what share of `dangerouslyDisableSandbox` calls followed a failed sandboxed run of the same command. The operating rule is to bypass only after a sandboxed run actually failed. Nothing in the index reported whether that held: `sandbox` lists individual bypasses, and `sandbox-bypass-effective-command` ranks verbs by raw frequency.

The query emits one row per (host, normalized verb) carrying `justified`, `reflexive`, `justified_pct`, and distinct sessions. Each host also gets an `(all)` rollup row that sorts first and spans every verb, including those below `min_count`.

The verb normalization is lifted from `sandbox-bypass-effective-command`, which strips leading `cd` wrappers, `echo` lines, and env-assignment prefixes. Reusing it keeps the two queries agreeing on what counts as one command. Checking the `local` rollup against a hand-written aggregate over `sandbox_bypasses` confirms the stripping drops no rows.

Measured against the live index, the justified share on the `local` host sits under 2%. The catalog entry states what that does not establish: a back-link needs a prior failure of the byte-identical command by the same agent in the same session. Failures learned in an earlier session, in another repo, or from a command differing by a path leave no trace. So read a low percentage as a verb worth checking for an `excludedCommands` entry, not as proof the bypass was unnecessary.


plugins/claude-code/skills/session/resources/queries/sandbox-bypass-justification.sql
```diff
@@ -0,0 +1,70 @@
...
+-- Params: after_date, before_date, project, host, min_count (floor on a verb's total,
+-- default 5).
+WITH stripped AS (
+  SELECT
+    sb.host,
+    sb.session_id,
+    sb.retried_tool_id,
+    regexp_replace(
+      trim(sb.command),
+      '^((cd [^&;\n]*|echo [^&;\n]*)(&&|;|\n)\s*|(export )?[A-Za-z_][A-Za-z0-9_]*=(\$\([^)]*\)|"[^"]*"|''[^'']*''|[^\s;&|]*)\s*(&&|;)?\s*)+',
```
bendrucker: pretty crazy regex, a comment would be helpful

bendrucker: Or I guess inline comments would be good as opposed to single monolithic, forget if Duck/SQL supports

bendrucker (commented):
LGTM, merge after breaking up comment

plugins/claude-code/skills/session/resources/queries/sandbox-bypass-justification.sql
bendrucker: Split the pattern into concatenated literals so each alternative carries its own comment. DuckDB folds them back into one constant, and the query returns identical rows.

```sql
      '^('
        || '(cd [^&;\n]*|echo [^&;\n]*)(&&|;|\n)\s*'     -- a `cd ~/src &&` or `echo hi;` wrapper
        || '|'                                           -- or
        || '(export )?[A-Za-z_][A-Za-z0-9_]*='           -- an env assignment, optionally exported
        || '(\$\([^)]*\)|"[^"]*"|''[^'']*''|[^\s;&|]*)'  -- whose value is `$(...)`, quoted, or bare
        || '\s*(&&|;)?\s*'                               -- and whose trailing `&&` or `;` is optional
      || ')+',
```

`sandbox-bypass-effective-command` carries the same regex, so it got the same treatment. The two blocks are byte-identical, which is what makes the header claim that both queries normalize alike hold.

PR #1239, opened 2026-08-21
statusline: keep color when stdout is piped

bendrucker:
Bun 1.4.0 changed `styleText` to suppress color when stdout is not a TTY. Claude Code always pipes the statusline, so both statusline scripts render entirely without escape codes on that runtime. Failing snapshots were the symptom, and the visible break is a monochrome statusline.

`validateStream: false` restores the codes at the call site. The new `user/scripts/style.ts` wraps `styleText` with that option so both scripts and the subagent test import one override rather than repeating it across every call site. `FORCE_COLOR` would work too, but it is process-global and every command the statusline spawns would inherit it.

`oven-sh/setup-bun` was unpinned, so CI floated onto 1.4.0 and turned red on an untouched `main` with nothing in the log connecting the failure to a runtime upgrade. Pinning it means the next Bun release shows up as a diff to review.

Verified against the piped-stdout path the client actually uses, which emits the same escape codes as it did before the upgrade. The existing snapshots cover the restored output unchanged, so none were rewritten to accommodate this.

### DIFF
--- a/.github/actions/bun-install/action.yml
+++ b/.github/actions/bun-install/action.yml
@@ -6,6 +6,11 @@ runs:
   steps:
     - name: Setup Bun
       uses: oven-sh/setup-bun@v2
+      with:
+        # Pinned so a Bun release lands as a reviewable diff. Unpinned, CI
+        # floats to the current release and a runtime behavior change arrives
+        # as an unexplained red build on an untouched branch.
+        bun-version: 1.4.0
 
     - name: Cache dependencies
       uses: actions/cache@v5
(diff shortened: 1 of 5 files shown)


bendrucker:
Folded into #1238. This closed itself as merged when `50daa784` landed on `md-code-fragment`, which was still its base from the stack attempt. The commit is unchanged and ships in #1238.

Both failures came from the same Bun 1.4.0 float, and neither branch could go green alone: each repaired exactly the required check the other failed. Stacking could not resolve that, since the bottom layer is always missing one of the two fixes.

Issue #1246, opened 2026-08-25
Adopt from dmmulroy and mattpocock skill collections


bendrucker:
Adoption of selected skills from two upstream collections, settled through a grilling session. Nine lanes, one PR each.

## Sources

`dmmulroy/.dotfiles@a6d5117` holds the live set of 23 skills under `home/.agents/skills/`. The public `dmmulroy/skills` repo is a stale 2026-07-23 snapshot of 7 skills and is not the source for anything here.

`mattpocock/skills@6654f6b` supplies `tdd` from `skills/engineering/`.

Licensing splits across the two. `mattpocock/skills` is MIT, and `write-discoverable-code` self-licenses MIT in its own frontmatter. `dmmulroy/.dotfiles` carries no LICENSE file, and `show-me`, `writing-for-agents`, and `diagnosing-bugs` exist only there. Those three get their ideas and structure taken and their prose written fresh.

## Shared constraints

- Use `LANGUAGE.md` vocabulary: module, interface, depth, seam, adapter, leverage, locality. Never "boundary", which upstream prose uses throughout.
- Adapt, don't vendor. No sync scripts, no do-not-edit markers. Credit source repo, path, sha, and license in a README line.
- A review angle carries only what a reviewer can detect in a diff. The paired skill carries the reasoning. An angle spawns as an isolated agent and cannot follow a pointer, so it gets no cross-reference.
- Write under the doctrine that lands in wave 0.
- Finish with `/ship`. Report the PR URL and anything that changes the plan for remaining lanes.
- Touch nothing outside your lane. The partition rests on file independence.

## Wave 0

Blocks wave 1. Five lanes author new skills, and this is the guidance on how to write them.

- [ ] `agent-writing-doctrine`: `claude-code:skill` gains `references/information-hierarchy.md` (context load vs cognitive load, the information hierarchy ladder, progressive disclosure, co-location, sprawl) and `references/levers.md` (completion criteria as clarity plus demand, premature completion, leading words, negation, pruning against single-source-of-truth and cache-vs-environment). `SKILL.md` gains a done-state clause sharpening "Don't Railroad Claude": name the observable done-state rather than adding a step. A criterion binds flat reference as readily as a sequence, so this introduces no step lists.

## Wave 1

Six lanes, parallel, no shared files.

[...]

PR #1258, opened 2026-08-25
herdr: cache what --help cannot supply

bendrucker:
Adds six agent-lifecycle and safety behaviors to the `herdr` skill and removes the passages that cached a `--help` lookup. Part of #1246.

The skill already states that `herdr <group> <command> --help` enumerates every enum flag and default, and that rule decides what belongs in the file. `agent_not_ready`, the bracketed-paste write, and the rejection of terminal IDs and bare agent-kind labels are invisible to `--help`. The `plugin pane open --placement` matrix, the `reviewr` contract, and the `pane wait-output` flag summary were not, so they are gone or behind a pointer.

Two decisions the diff will not show. The placement note keeps one clause because `--help` lists four placements while the binary also parses `popup` and `fullscreen`. Cutting the `agent prompt` timeout paragraph made two `--help` lookups the discovery path where prose used to sit, hence the two new `allowed-tools` entries.

`SKILL.md` lands about 810 tokens under `main` with all six behaviors in place.

`agent_not_ready` and `agent_blocked` were observed in [dmmulroy/.dotfiles](https://github.com/dmmulroy/.dotfiles/blob/a6d5117/home/.agents/skills/herdr/SKILL.md) at sha `a6d5117`. The prose here is original.


user/skills/herdr/README.md
bendrucker, line 1: Excessive narration slim way down

bendrucker: Deleted the file. Every line in it was deliberation, and nothing loaded or referenced it. Attribution moved to the commit message and the PR body.

Applied the same test to the two files that ship, cutting sentences that argue a rule is correct or name what it replaced:

- The `Command Surface` preamble announcing where the block comes from.
- The `plugin action list` example in `Output Formats`, derivable from the list above it.
- The paragraph explaining why `agent prompt` needs `--wait`. The instruction to drop it only for unattended work stays.
- The clause justifying the `--placement` note and the one explaining Windows action variants, both trimmed to the fact that changes what you do.
- In `reviewr.md`, the "rules it out as the explanation for an unexpected diff" clause, and the reasoning behind snippet-as-anchor and the single-copy batch.

`SKILL.md` is at ~3,776 tokens, 725 under `main`. The tree drops the README's ~200 on top of that.

PR #1272, opened 2026-08-27
lint: triage and enforce no-await-in-loop

bendrucker:
`eslint/no-await-in-loop` is now enforced at error. Each of the 147 sites it flags took one of two outcomes: the loop was rewritten to run its iterations concurrently, or it stays sequential and carries an inline `oxlint-disable-next-line` naming the constraint that forces the ordering. No disable is bare.

The reasons are the review surface. A rewrite that reorders results and a disable whose stated reason does not hold are the two ways this can ship wrong, so both tables below name a representative site you can check against the code.

### Kept Sequential (108)

| Reason class | Sites | Representative |
|---|---|---|
| Shared mutable state | 40 | `session/scripts/db.ts:354`: `SET VARIABLE` is connection-global state the import below reads back |
| Iteration depends on the previous | 23 | `github/scripts/pr-comments.ts:326`: each page's cursor comes from the response before it |
| Retry / backoff | 10 | `mac/scripts/jxa.ts:130`: sleeps before retrying a transient Apple Events failure |
| Rate limit / remote pacing | 8 | `things/src/mcp/tools.ts:598`: the Things URL scheme drops batches sent back to back |
| Ordered output | 8 | `packages/validate/run.ts:41`: each file prints its bullet as it is validated |
| Poll loop | 7 | `review/skills/inbox/scripts/watch.ts:56`: poll interval between iterations |
| Bounded tiny N | 8 | `session/scripts/refresh.ts:51`: two known stray directories |
| Worker-pool body | 2 | `evals/pr-body/scripts/run-eval.ts:241`: the pool above is the parallelism |
| Memory bound | 1 | `writing/skills/scan/scripts/scan.ts:73`: the audit runs over an arbitrary tree |
| Bounded subprocess fan-out | 1 | `comments/skills/audit/scripts/audit.ts:307`: one `sh -c` formatter per edited file |

Shared mutable state resolves to six shared resources:

- one DuckDB connection, which `session/scripts/db.ts` alone holds 13 sites against
- one `GIT_INDEX_FILE` in `comments/apply/branch.ts`
- one `FileSink` in the eval writers
- one launchd user domain in `scheduled.ts`
- mermaid's global parser
- `COPILOT_HOME`'s session ledger

### Parallelized (39)

| Group | Loops | Sites | Example |
|---|---|---|---|
| Eval corpus and label readers | 10 | 15 | `evals/pr-body/scripts/run-eval.ts` `loadScenarios` reads every scenario JSON at once |
| Repo and config scanners | 9 | 13 | `scripts/check-plugin-deps.ts` scans each plugin's package.json and imports concurrently |
| Hooks and statusline | 2 | 4 | `user/scripts/subagent-statusline.ts` reads each subagent's meta and transcript concurrently |
| Tests | 7 | 7 | `things/src/mcp/stdout.test.ts` reads the import closure concurrently |

Every rewrite keeps result ordering. `Promise.all` over `map` preserves index order, and the collecting loops end in `flat()` or an ordered reduce. No rewrite touches a loop that spawns a subprocess with side effects, drives an interactive or long-lived loop, or writes shared state.

### Verification

None of the rewritten check scripts has unit coverage that would catch an argument-shape break, so each ran directly against the repo. `check-permission-rules`, `check-plugin-deps`, `validate/hooks.ts`, and `skill-lint "plugins/writing/skills/*"` all produce their expected output.

The failures this repo already carries on `main`, in `buildAttribution`, `upstream.ts`, `readRemoteUrls`, and the session index, sit in paths this branch does not touch.

### Config

This branch rebases onto every sibling oxlint PR, so `.oxlintrc.json` now carries all of their rules. This change adds the one `no-await-in-loop` key on top of them.

### DIFF
--- a/.claude/hooks/ox/index.test.ts
+++ b/.claude/hooks/ox/index.test.ts
@@ -189,9 +189,11 @@ describe("ox hook", () => {
   // wrote by name: the directory holds live sessions' counts too.
   afterAll(async () => {
     await rm(tempDir, { recursive: true, force: true });
-    for (const session of gateSessions) {
-      await rm(blockCountPath(session), { recursive: true, force: true });
-    }
+    await Promise.all(
+      [...gateSessions].map((session) =>
+        rm(blockCountPath(session), { recursive: true, force: true }),
+      ),
+    );
   });
 
   describe("runOxlintAgent", () => {
@@ -430,6 +432,7 @@ describe("ox hook", () => {
 
       const budget: (SyncHookJSONOutput | null)[] = [];
       for (let stop = 0; stop < STOP_BLOCK_LIMIT; stop++) {
+        // oxlint-disable-next-line no-await-in-loop -- each stop increments the persisted block count that the next stop reads.
         budget.push(await processStop(mockStopHookInput(transcript, stop > 0, session)));
       }
       const past = await processStop(mockStopHookInput(transcript, true, session));
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -7,6 +7,7 @@
   },
   "rules": {
     "prefer-template": "error",
+    "no-await-in-loop": "error",
     "no-template-curly-in-string": "warn",
     "typescript/no-non-null-assertion": "error",
     "typescript/no-unnecessary-type-assertion": "error",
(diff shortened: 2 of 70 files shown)


bendrucker:
@greptileai review

PR #1294, opened 2026-08-28
review: serve the review plan from a CLI

bendrucker:
`review:code` opened Phase 1 by sending the model to `efforts.md` to derive its own review shape from prose: cross a family row with a level column, look the cell up in a 9-row budget table, then filter fourteen angles across two more files down to that cell's set. 125 of the 149 sessions that invoked the skill pulled `efforts.md` into context to do it, several after failed `sed` and `grep` extractions, because the cells are table rows with nothing to anchor a range on.

Phase 1 is now one Bash call. `scripts/review-plan.ts` resolves the cell from the model id and the effort token, then prints the mode, caps, verify rule, sweep flag, framing paragraph, and the text of exactly the angles that cell runs. SKILL.md tells the review to follow the printed plan and not to open the data files itself.

## Data Stays Data

The cells live in `efforts.yaml` and the angle text in `angles.yaml`, each with a draft 2020-12 schema beside it and a `yaml-language-server` directive pointing at it. No cell name, cap, framing, angle, or instruction paragraph is a literal in the TypeScript. The renderer walks the resolved cell and prints what it finds.

YAML rather than JSON so multi-paragraph prose sits in `|` block scalars and diffs line by line. The schemas close every vocabulary with `additionalProperties: false`, so a typo in a cell name fails in the editor and against `ajv`, instead of silently producing a review with the wrong shape.

`angles.yaml` was generated from the two markdown files by script, and the reconstruction diffed byte-for-byte against their `HEAD` versions. `efforts.md` keeps only the reasoning a human needs before changing the cells.

## Level Tokens

The plan called for a fixed stem list (`low`, `med`, `hig`, `xhi`, `max`). That cannot satisfy its own cases. `hi` and `xh` are shorter than the stems and must resolve, while `mediumish` matches the `med` stem and must not. `resolveLevel` matches by unique prefix over the level names plus the aliases instead, with exact matches winning outright and ambiguity reported rather than silently picked.

## Coverage Gap

`writing:scan audit` no longer reaches this prose. `plugins/writing/detection/paths.ts` limits `PROSE_EXTENSIONS` to `md`, `markdown`, `txt`, `mdx`, `rst`, and `adoc`, so moving angle and framing text into `.yaml` drops it out of trope auditing. Adding `.yaml` there is a follow-up.

Editor-side schema validation is unproven. The repo has no `.vscode/settings.json` and these are its only two `yaml-language-server` directives, so nothing headless exercises the extension's relative `$schema` resolution.


bendrucker:
@greptileai review

PR #1300, opened 2026-08-29
github: name the run that actually failed

bendrucker:
The reported symptom was that the watcher counted skipped and cancelled runs as failures. It didn't. Every failing event on #1298 was a real verdict. The bug was in what the event *pointed at*: `probePr` derived state from `gh pr checks`, then took the run id from `gh run list --branch <name> --limit 1`. That is the branch's newest run, related to the failing check only by coincidence, so `github:logs` kept fetching runs with no failing jobs.

`selectRunId` now reads the runs on the failing commit, keeps only conclusions that leave job output behind (`failure`, `timed_out`), and collapses each workflow to its newest run so a failure a re-run replaced is not named again. Replaying 779adfd9 through it yields 33220042315, the real `lint` failure, where the old path yielded the skipped 33220042301.

Only a skipped check is neutral. GitHub counts a skipped required check as passing; branch protection does not accept a cancelled one. Folding cancel into the same bucket exits the watcher green on a PR that cannot merge, so a cancel holds the target at `running` and waits out the wall clock instead.

A failing event can now carry `"run_id": null`, meaning the red check is an external status with no logs to fetch. The monitor skill, `github:logs`, and `babysit` each document that case.

The attribution query decodes through a zod schema, following the convention that landed on main while this branch was open.

`deriveRunListState` is untouched, so branch and run-id mode still map `cancelled` to failing and `skipped` to success. Those watch runs rather than checks, and the event vocabulary has no way to say terminal but inconclusive.

### DIFF
--- a/plugins/github/agents/logs.md
+++ b/plugins/github/agents/logs.md
@@ -51,3 +51,4 @@ Keep `lines` short (10 to 50 lines per job). The caller can read `log_file` for 
 
 - Do not analyze root causes or suggest fixes. Extract and summarize.
 - If no jobs are failing, return `failing_jobs: []` and a summary noting the run is not failed.
+- If the run id is missing or null, no Actions run failed. Return `failing_jobs: []` with that summary and run no `gh run view`.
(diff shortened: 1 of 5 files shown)


bendrucker:
@greptileai review

PR #1356, opened 2026-09-05
lint: enable five oxlint collection rules

bendrucker:
Adds oxlint's five array and object collection rules to `lint/base.json`, each with a comment on what the banned form costs:

- `unicorn/no-array-for-each`
- `unicorn/explicit-length-check`
- `unicorn/prefer-spread`
- `prefer-object-spread`
- `unicorn/prefer-math-min-max`

Fixes every violation this surfaced:

- 20 `forEach` callbacks converted to `for...of`
- one of those relied on the callback's own `return` as a continue, converted to `continue`
- one `prefer-spread` survivor
- auto-fixed hits for `explicit-length-check`, `prefer-object-spread`, and `prefer-math-min-max`

Two sites keep the banned form behind an inline `oxlint-disable-next-line`. The preferred form collides with an existing, correctly firing rule.

Spreading a plain ASCII string trips `typescript/no-misused-spread`'s Unicode code-point warning in `scripts/coverage/lcov.test.ts` and `plugins/pull-request/scripts/prose.test.ts`.

A fast-check `.map()` callback in `plugins/comments/detection/rank.test.ts` needs a fresh copy. `oxc/no-map-spread` reads that copy as an inefficient in-place mutation.

`oxlint --fix-suggestions` introduced a bug while auto-fixing the `prefer-object-spread` hit in `rank.test.ts`. It rewrote `Object.assign({}, c, { path })` into `Object.assign(c, { path })`. That mutates the source comment in place instead of copying it.

Replaced with an object spread. The five other `Object.assign` call sites in the repo mutate an existing object. The rule only flags the fresh-literal form, so they stay as they are.

### DIFF
--- a/lint/base.json
+++ b/lint/base.json
@@ -26,14 +26,35 @@
         "ignoreRestSiblings": true
       }
     ],
+    // Object.assign({}, ...) merges through a function call instead of an
+    // object literal, hiding the resulting shape from tools that read
+    // literals better than call expressions, and needs an empty target
+    // argument purely to avoid mutating an input.
+    "prefer-object-spread": "error",
     "prefer-template": "error",
     "import/no-named-as-default-member": "error",
     "oxc/no-map-spread": "error",
     // A `new Promise` executor that only wraps an existing async API duplicates
     // logic the platform already exposes, and drops its built-in error handling.
     "promise/avoid-new": "error",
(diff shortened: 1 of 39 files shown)


bendrucker:
@greptileai review

PR #1359, opened 2026-09-05
lint: enable import-hygiene rules

bendrucker:
Enables three import-hygiene oxlint rules and fixes every violation they surface.

- `import/no-namespace` bans `import * as x from "mod"`. A namespace binding hides which members a file actually uses, from both a reader scanning the imports and the unused-import check.
- `no-duplicate-imports` bans two import statements pulling from the same module, which invites the bindings to drift out of sync.
- `unicorn/prefer-export-from` bans importing a name into local scope for the sole purpose of re-exporting it.

`import/no-namespace` ignores `fast-check`, whose documented idiom is `import * as fc from "fast-check"`.
The other namespace imports (`node:path`, `node:os`, `acorn`, and two local modules) had no equivalent justification, so they're converted to named imports with every call site rewritten to match.
`no-duplicate-imports` violations are merged into one statement per module, using an inline `type` specifier for the type-only bindings.
`unicorn/prefer-export-from`'s three sites were fixed automatically by oxlint's `--fix-suggestions`.

`bun run check` and `bun test` both pass.

### DIFF
--- a/evals/scripts/collect-ci-runs.ts
+++ b/evals/scripts/collect-ci-runs.ts
@@ -3,7 +3,7 @@
 import { readdirSync } from "node:fs";
 import { mkdtemp, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
-import * as path from "node:path";
+import { join } from "node:path";
 import { cli } from "cleye";
 import { z } from "zod";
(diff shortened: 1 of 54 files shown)


bendrucker:
@greptileai review

PR #1361, opened 2026-09-06
comments: gate the rubric on the judge that ships

bendrucker:
The eval gate pinned its SDK oracle to `claude-sonnet-4-6` while `judge.workflow.js` runs the judging agents on the `sonnet` alias, so the gate certified a generation production never uses. It also read `ANTHROPIC_API_KEY` through `op run`, which needs someone at a 1Password prompt. That is how #1348 rewrote the rubric and merged without the gate running at all.

`eval.ts build` now shards the labeled corpus through the same writer the audit's `preflight` uses and prints a `<preflight>` block. Hand that to the Workflow tool, then `eval.ts score --job <dir> --gate` scores the verdicts the agents wrote. That path runs on subscription auth.

## Gate of Record

The workflow path gates, because it exercises the code that ships and runs unattended.

The oracle is a keyed cross-check and is no longer a calibration reference. Models after Opus 4.6 reject `temperature` with a 400, confirmed against the installed SDK's `@deprecated` annotation, so the batch pins `output_config.effort` and samples. Two runs over one corpus can disagree. `max_tokens` rose to 16,000 because Sonnet 5's adaptive thinking draws on the output budget and truncated the verdict JSON at the old 4,096.

`evals/op.env` held an item ID in the vault slot of its `op://<vault>/<item>/<field>` reference, breaking every keyed runner in the repo. The item's title contains the path separator, so it is addressed by ID.

## Results

The workflow gate scores the corpus at accuracy 1.00 with zero keep violations.

The oracle scores it at 0.96, also with zero keep violations. Its one disagreement reads `db-host-variable-voice-rewrite` as `keep` where the label says `rewrite`, which leaves a comment's voice alone rather than destroying a justified one. The gate keys on keep violations for that reason: an over-cautious judge is a worse eval score and a safe apply.

An oracle batch twice died parsing an unterminated string without the `max_tokens` guard firing, then passed on a retry. The parse failure now carries the stop reason, the output tokens against the budget, and the tail of the response, so a third occurrence names its cause.

## Limits

No CI check requires the gate. The workflow path needs the Workflow tool that only a Claude Code session provides, so a rubric change can still merge unvetted.

The oracle's `claude-sonnet-5` pin and the workflow's `sonnet` alias drift apart on their own when the alias moves.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -57,12 +57,13 @@ Per-skill harnesses live inside the plugin they measure, at `plugins/<plugin>/ev
 - `bun plugins/issue/evals/issue-refine/scripts/build-dataset.ts`, then `label/server.ts`, then `scripts/ab-report.ts` and `scripts/judge.ts`
 - `bun plugins/review/evals/review-voice/scripts/mine.ts`, then `label/server.ts`, then `scripts/report.ts`
 - `bun plugins/writing/evals/writing/scripts/mine.ts`, then `label/server.ts` (scorer and judge are not built yet)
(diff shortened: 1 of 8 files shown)


bendrucker:
@greptileai review

PR #1362, opened 2026-09-06
lint: scope vitest/prefer-each to test files

bendrucker:
`vitest/prefer-each` rewrites a loop over cases into `it.each`, which only means something when the loop body is a test. #1355 enabled it repo-wide, where it fires on any `for...of`. It broke main on meme's `render.ts`, whose loop draws one canvas line per iteration and has nothing to do with testing.

This is the hazard `vitest/no-commented-out-tests` already documents two lines above, so it takes the same fix: off at the top level, on in the `**/*.test.ts` override. A probe test file with a real loop-of-tests still gets flagged, so the rule keeps working where it applies.

### DIFF
--- a/lint/base.json
+++ b/lint/base.json
@@ -68,17 +68,21 @@
     // `it(...)` call but isn't one. Off here. Enabled only under
     // **/*.test.ts below, where a commented-out test can actually occur.
     "vitest/no-commented-out-tests": "off",
-    // This repo tests with bun:test, not vitest. The vitest plugin matches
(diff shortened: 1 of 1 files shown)


bendrucker:
Superseded by #1354, which landed the same scoping (base-level `vitest/prefer-each: "off"` plus the `**/*.test.ts` override). Rebasing this branch onto main leaves an empty diff — the only remaining delta was a reword of the explanatory comment.

PR #1398, opened 2026-09-09
skills: add grill-docs

bendrucker:
`grilling` keeps everything it settles in the session window, so a compaction or an abandoned session takes the vocabulary and the decisions with it. `grill-docs` runs the same interview and writes each result the moment it resolves.

Vocabulary goes to `.claude/rules/<area>.md` under a `paths` entry, or to `CLAUDE.md` when the term governs the whole repo. Both surfaces reload on their own, which a root glossary file does not. Decisions go beside the plan that produced them.

A decision earns a record only when reversing it costs real work, it would surprise someone arriving without the conversation, and something real was given up to make it. Most sessions clear none of the three, which keeps the file small enough to stay worth reading.

The interview itself stays in `grilling`. Copying the body would fork the round rules, and injecting it would carry `grilling`'s own bang-injection of `ask` through unexecuted.

Typed entry only, matching how `grilling` is reached when the user names it. That costs no recurring context, so `/skill-config-vs-observed` reporting zero `calls` and zero `typed` after a few weeks is the signal to delete it.

### DIFF
--- a/user/skills/grill-docs/SKILL.md
+++ b/user/skills/grill-docs/SKILL.md
@@ -1,0 +1,56 @@
+---
+name: grill-docs
+description: Run the grilling interview and write what it settles to the repo as it settles, so the vocabulary and the decisions outlive the session window.
+argument-hint: "[<what to grill>]"
+disable-model-invocation: true
+allowed-tools:
+  - AskUserQuestion
+  - Agent
+  - Read
+  - Write
+  - Edit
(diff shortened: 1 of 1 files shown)


bendrucker:
Holding this for a wording pass before merge. Use less jargon and lean on known ideas like tradeoffs instead of phrasings like "Something real was given up to make it".

PR #1421, opened 2026-09-12
chief: add the alert-triage daemon and agent

bendrucker:
Adds `chief`, one daemon plus one Sonnet agent that triages every alert other Claude sessions raise so only what needs you gets through. Today `moshi-hook` and `vibe-island` push everything with no tiering. This plugin puts every hook event into an append-only ledger, tiers it deterministically (`now`, `boundary`, `digest`), and rings a persistent herdr pane running the chief agent whenever a row is due. The agent holds or acks with a one-line reason. Nothing else pushes.

## Shape

The daemon is `Bun.serve` on `127.0.0.1:7391` with `/mcp`, `/node/mcp` (append and status only, for a later work-Mac node), `/ingest` (loopback only), and `/healthz`. The plugin's stdio MCP entry ensures the daemon is up and bridges to it, so it starts when the chief pane opens and outlives it. No launchd entry. The plugin stays out of `enabledPlugins`; a settings file in `~/src/chief/` enables it for that one session.

Presence is a stub returning nulls, so every boundary is the top of the next hour. The ntfy path is wired but dormant until an ntfy server exists, since Homebrew's macOS `ntfy` is a client-only build. The dotfiles side is a separate PR.

## Decisions

- The MCP endpoint builds a server and transport per request. A stateful transport binds to its first client and rejects every later initialize, and the daemon serves many bridges over its life.
- The doorbell text is `/chief:chief drain`. A plugin skill carries the plugin prefix, and `/chief drain` returned Unknown command in the pane.
- `disallowedTools` on the agent lists tool names only. `Bash(git push:*)` there removed Bash entirely.
- `herdr agent list` wraps its result in an envelope and names the pane `pane_id`. Pane resolution degrades to no pane when herdr is unavailable.

## Verification

Started from the README's pane command with `--plugin-dir` at this checkout. The bridge spawned the daemon, the pane listed the `mcp__plugin_chief_chief__` tools, and the daemon answered `/healthz` after the session exited. Synthetic payloads through the hook tap produced a `boundary` permission row resolved by `SessionStart`, an `ask_user` row releasing at the top of the hour, and a credential row that rang `/chief:chief drain` in the pane within a minute and was acked into `decisions.jsonl`. A probe daemon with a one-minute tick sent `/flock tick`. Added tests against a real `ntfy serve` covering publish, the Hold 1h reply, and reconnect.

Deferred: the presence readers, retiring the `moshi-hook` and `vibe-island` entries, an ntfy server strategy for macOS, and the tailgate upstream entries in the live config.


bendrucker:
Copilot one-shot review (terra, 27 credits) raised five findings. Fixed in bc997968: the spool drain race, overlapping release passes, and a non-2xx ntfy publish counted as delivered, plus the config schema requiring an ntfy block the README says can be absent. Rejected: pane resolution swallowing a failed herdr agent list. A hook event must land in the ledger even when herdr is down, and the row's missing pane is the visible signal.

bendrucker:
Closing: chief was an experiment in a single assistant session with schedule awareness across personal tools. Declared a failed experiment on 2026-09-15. Branches removed, transcripts kept.

PR #1428, opened 2026-09-12
stop: install only where there is a manifest

bendrucker:
`bun install --cwd <cwd>` ran inside the same `try` that wraps prek, so a session stopping in a directory with no `package.json` failed on the install and fell into the catch. That catch formats everything as `Check failures:`, which meant Stop blocked with a prek error for a prek run that never happened, every time, with nothing the author could fix in their code.

The install moves ahead of the `try` and is skipped when the directory has no manifest. A failed install no longer blocks either: prek reports a missing dependency itself when it actually needs one, so the install is a convenience and has no business deciding whether the session can stop.

Both tests stub `bun` and `prek` onto `PATH` and assert which commands got invoked, since the behavior at stake is what runs rather than what it returns.

Found while this Stop hook blocked a session, so there is no Things todo behind it.

### DIFF
--- a/.claude/hooks/stop/index.test.ts
+++ b/.claude/hooks/stop/index.test.ts
@@ -144,6 +144,55 @@ describe("processStop", () => {
     expect(await processStop(stopInput({ transcript_path: transcriptPath }))).toBeNull();
   });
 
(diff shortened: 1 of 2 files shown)


bendrucker:
@greptileai review

PR #1443, opened 2026-09-15
user: deflake the pane metadata report test

bendrucker:
A full-suite run on 2026-09-06 failed once without recording which test, and the pass count differed between runs. Ten instrumented runs settled both questions.

The count never varied, ruling out registration. f2accdf2 had added a test to `corpus-selection.test.ts` between the two runs being compared. The one failure was `pane metadata report > carries the title token and the dial to herdr`, asserting `[]` against the expected token list.

Two independent races produce that assertion.

The stub appends one line per argument, and the poll returns as soon as the log contains `report-metadata`, which is argv[1] of about twenty. A reader arriving mid-loop sees a record whose tokens are not there yet. The stub now builds the record in a `.part` file and renames it into place.

The report cache lives under `tmpdir()` keyed by session id, and the id was hardcoded. Any concurrent or interrupted run on the machine could leave an entry whose signature matched. `shouldReport` then returns false, the child never spawns, and the poll waits out its full budget against a stub nothing called. Each call now generates its own id.

The failing run's own `finally` deletes that cache entry. That is why a retry was always green, and why this never reproduced on demand.

Stress runs isolate the two. The partial read measured one failure in two hundred runs before the change and none in four hundred after. For the cache, planting an entry with a matching signature failed the test every time before and does nothing after.

Review turned up a third defect in the same lines. The log path interpolates into the generated `/bin/sh` unquoted, so a `tmpdir()` containing a space word-splits the redirect target and the `mv` arguments. Nothing checks the stub's exit code and the script carries no `set -e`, which surfaces the failure as that same timeout error.

6fe05e87 hardened this test's timeout last week and described the symptom as `expect([]).toEqual(["title=..."])`, the same shape. It addressed neither cause.

Separately, `origin/main` is red for an unrelated reason. 070a31aa changed `plugins/writing/linguistics/tags.ts` without updating the `plugins/pull-request/scripts/linguistics/tags.ts` copy that the drift guard requires to be byte-identical. The plugin matrix is path-filtered and this branch touches no plugin, so that job does not run here. It still needs its own fix.


bendrucker:
@greptileai review

Issue #790, opened 2026-06-10
writing: cross-sentence detector layer


bendrucker:
The cross-sentence layer is defined in `linguistics.md` with no tooling. The existing coverage is one regex (`CROSS_SENTENCE_NOT`, 8 hits in 2 sessions over 30 days, confirming it catches almost nothing) plus `connectorDensityHits`. The 2026-06-09 discourse-detection survey yields five buildable detectors, none requiring a new dependency. All are batch-surface detectors: they need multi-sentence windows and the recall-leaning bar.

## Shared Infrastructure First

Extract `splitSentences` from `connectorDensityHits` into a shared utility and add `splitParagraphs` (split on blank lines, then sentences per paragraph). Every detector below consumes these. The one tagger-importing detector lives in a module hooks never import (same wall as `linguistics/classifiers.ts`).

## Build Order

1. Adversative negation-flip: negation cue in sentence N (`not`, contracted negations), adversative opener in N+1 (`However,`, `But`, `Yet`, `Nevertheless,`). Pure regex over a two-sentence window, no coreference needed because the marker itself signals the flip. Completes the contrast family alongside the within-sentence patterns.
2. Burstiness: coefficient of variation of sentence word counts. The literature places default LLM output near CV 0.5 and human prose near 0.78, with the caveat that the thresholds come partly from commercial detector docs. Gate on 8+ sentences. Pure arithmetic.
3. Question-then-answer cadence: flag when 2 of 4+ paragraphs open with a question. Pure regex, lower precision, near-zero cost.
4. Tricolon detection: normalized edit distance over coarse POS tag sequences (filtering DET/PUNCT/NUM) across consecutive sentence triples, threshold near 0.35 with each sentence 6+ tokens. Uses the existing compromise adapter, so batch-only, and the one detector requiring a labeled eval pass per the #769 protocol before shipping.
5. Discourse-marker density: generalize the connector-density check from punctuation connectors to a two-tier PDTB-derived connective table (unambiguous adversatives as flat lookups; ambiguous ones gated on context).

## Constraints

- Every literature threshold is a starting heuristic. Calibrate against the session corpus before any rule fires, and record the calibration as the rule's earning evidence per #787.
- Burstiness-family features have documented false-positive exposure on terse technical prose, so they stay length-gated and batch-only. None of these five is a hook candidate in its first iteration.

Refs #769, #787.

PR #1444, opened 2026-09-15
writing: gate four structural metrics on the null

bendrucker:
Adds the four structural rate features from #790 to `voice-delta.ts` and puts each through the permutation null. One survived.

Corpus A averages 295 words per document against corpus B's 92, so a feature that varies with length clears its floor on that difference alone. `--min-words` and `--max-words` hold both corpora to one band. Held to 100-400 words, 15 of the 16 shipped features still clear and `body_length` itself drops from 17.25x to 2.53x, which is the control working rather than flattening everything.

| candidate | full corpus | 100-400 words | disposition |
|---|---|---|---|
| discourse marker density | 1.63x | 1.41x | shipped |
| sentence-length burstiness | 1.37x | 0.35x | retired, length artifact |
| paragraph-length uniformity | 0.21x | 0.90x | retired, below floor everywhere |
| parallel-construction rate | 0.38x | 0.80x | retired, below floor everywhere |

Gap over floor at 2,000 splits, seed 1. The corpora are local-only, so these reproduce on your machine and not in CI.

Block structure comes from `mdast-util-from-markdown`, which changed a result rather than only satisfying `.claude/rules/scripts.md`: splitting paragraphs on blank lines counted every heading as a one-word paragraph, and paragraph uniformity read 2.03x that way.

The retirements stay in `RETIRED_FEATURES`, scored by `--candidates` and read by nothing else, so the table stays falsifiable.


bendrucker:
@greptileai review

bendrucker:
@greptileai review

PR #1445, opened 2026-09-16
ci: select a plugin from paths it declares

bendrucker:
`origin/main` was red: #1440 edited `plugins/writing/linguistics/tags.ts` and left the pull-request plugin's checked copy behind, so the drift test failed only after the merge.

The copy exists because plugins are distributed one at a time and runtime code cannot resolve a `workspace:*` package, and the drift test is what keeps it honest. The test lives in the pull-request plugin's scope, and the matrix runs only the plugins a PR's changed files name, so a writing-side edit never ran the check that guards its copies.

## Selection

A plugin's `.ci.json` now takes a `paths` array of prefixes outside the plugin that also select it, and pull-request claims the two writing paths it copies from. The alternative was widening the workflow's `--always` list, which would have run all 40-odd plugin jobs on every writing edit to catch a dependency that only one plugin has.

`plugins/writing/hooks/heading-case.test.ts` already guards one of the four copies from the writing side, which is why a `heading-case.ts` edit would have been caught and a `linguistics/` edit was not. Mirroring the whole copy table over there would mean maintaining two lists instead of one, and it would still only assert byte equality. Selecting the pull-request job runs its whole suite, so a writing change that breaks a consumer of the copy fails too.

The `paths` list is itself something that can fall behind the copy table, so a test in the drift suite requires every copy source to sit under a declared path. Adding a fifth copied file without extending `paths` now fails in the same suite as the copy itself, which the edit to the table already selects.

`select` came out of the script's `main` as a pure function so the selection rules are covered directly: a plugin's own files, a claimed path, a sibling path nobody claims, and an `--always` hit.


bendrucker:
@greptileai review

PR #1446, opened 2026-09-16
oxlint: lint every rule case in one spawn

bendrucker:
The `packages/oxlint-rules` suite spawned oxlint once per case, so the first test paid the binary's cold start while every later test ran warm. On the runner that failed #1440 that first case took 5055ms against a 5000ms per-test timeout, the second took 1956ms, and the rest settled at ~150ms. A re-run of the same commit put that first case at 439ms. Nothing about the code changed between them, so no timeout value is safe: the cost is runner variance, and a larger number just moves the cliff onto a slower machine later.

Batching all 45 cases into a single oxlint run at module load takes process startup out of the tests entirely. Module evaluation carries no test or hook timeout, so a slow spawn now costs wall clock instead of a failure. Locally the suite drops from 7.1s to 0.2s, and the test names are unchanged.

The `DecodeError: invalid JSON: Unexpected EOF` in that log was downstream of the timeout. Bun killed the child ("killed 1 dangling process"), `spawnSync` returned empty stdout, and `decodeJson` parsed the empty string. A spawn that writes nothing now reports the exit code, the signal, and stderr, which distinguishes a kill from a crashed plugin load.

## Verification

A shim in place of `node_modules/.bin/oxlint` that delays the first invocation reproduced the CI failure byte for byte, down to the `killed 1 dangling process` line. Delaying it 3s instead of 6s passed, which rules out slow output being lossy on its own and pins the empty stdout to the kill. Against the batched suite the same 6s shim passes, and so does a 30s one.

Mutating a rule so it stops firing fails exactly the cases that cover it, so the per-rule filtering that batching requires has not made the assertions vacuous. Two guards back that up: duplicate case names are rejected rather than silently merged, and a lookup that finds no result throws instead of returning an empty array that every `allows` case would accept.

Enabling all five rules in one config is what lets the cases share a run. Each assertion filters to its own rule, which is the same question the old per-rule config asked.


bendrucker:
@greptileai review

bendrucker:
Added a 120s process timeout on the batched spawn. That is a hang backstop rather than a cost budget: the worst cold start observed was ~5s, so nothing about runner speed brings it near the limit, and the empty-output guard turns a SIGTERM into a named failure instead of a hung job.

@greptileai review

PR #1463, opened 2026-09-18
herdr: show a dispatched thread's tags on its pane

bendrucker:
A dispatched thread's tags are routing keys, and reading them meant running `ledger.ts status`. The agent sidebar already renders pane tokens, so `dispatch.ts` now stamps one token per tag on the new agent's root pane under a `dispatch` source, right after the agent binds its name.

Tokens are display only. A refused `report-metadata` call warns on stderr, and the dispatch and its ledger row go through unchanged.

Every tag rides in one call that herdr validates as a whole. A ledger tag key may be any length in lowercase, while a token name takes at most 32 characters of either case. A key herdr refuses would have cost the pane every other tag on the thread, so those keys are left out with their own warning.

Clearing the tokens when a thread closes belongs with `ledger.ts outcome` and is not here.

Tests drive the fake runner's recorded call sequence. The stamp carries every tag and lands after the agent start. An unnameable key drops out while its siblings stamp, and a refused stamp leaves the record intact.


bendrucker:
Folded into #1473 for a live trial.

